PR #4, opened 2019-03-04
Detect benchmarks that return promise and handle them as deferred ben…

qwwdfsad:
…chmarks

### DIFF
--- a/runtime/jsMain/src/org/jetbrains/gradle/benchmarks/js/JsBenchmarkSuite.kt
+++ b/runtime/jsMain/src/org/jetbrains/gradle/benchmarks/js/JsBenchmarkSuite.kt
@@ -2,12 +2,13 @@ package org.jetbrains.gradle.benchmarks.js
 
 import kotlinx.cli.*
 import org.jetbrains.gradle.benchmarks.*
+import kotlin.js.*
 import kotlin.math.*
 
 external fun require(module: String): dynamic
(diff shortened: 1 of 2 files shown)


runtime/jsMain/src/org/jetbrains/gradle/benchmarks/js/JsBenchmarkSuite.kt
```diff
@@ -87,7 +104,6 @@ class Suite(val title: String, @Suppress("UNUSED_PARAMETER") dummy_args: Array<o
                 else -> throw UnsupportedOperationException("Format ${args.traceFormat} is not supported.")
             }
 
-
             results.add(result)
```
qwwdfsad: > "  ~ ${score.format(d)} ops/sec ±${(error / score * 100).format(2)}%"

Not sure this is the best format for microbenchmarks.
E.g. this does not look readable to me:
```
… test.JsAsyncBenchmarks.promiseDelayedBaseline
    ~ 9.7,183 ops/sec ±0.52%

… test.JsAsyncBenchmarks.promiseResolve
    ~ 9,999,505 ops/sec ±2.87%

… test.JsTestBenchmark.sqrtBenchmark
    ~ 889,280,184 ops/sec ±1.15%
```

Maybe we should autodetect large numbers and change units automatically?
And detect fixed point in small number, e.g. `9.7,183` -> `9.718`

PR #10, opened 2019-09-10
Implement level configuration at which fixture method should run

qurbonzoda:


boswelja:
Is this being blocked by something?

PR #11, opened 2019-09-11
Implement Blackhole

qurbonzoda:


ilya-g:
To clarify: currently `Blackhole` is implemented as no-op in JS and Native just to allow use real JMH Blackhole in JVM from common benchmark code.

ilya-g:
Yes, we're struggling with implementing Blackhole for JS and Native platforms at the moment: we don't have enough time and expertise. 

Meantime providing a no-op blackhole in these platforms allows to reuse JVM benchmark code.

We hope to improve blackhole implementation later, and a PR with the improvement is welcome.

PR #12, opened 2019-10-10
Generate benchmark suites in their packages

qurbonzoda:


examples/kotlin-multiplatform/src/commonMain/kotlin/another/CommonBenchmark.kt
```diff
@@ -0,0 +1,22 @@
+package test.another
```
qurbonzoda: Perhaps this file should be removed before merging

PR #31, opened 2021-02-09
Update Gradle, update Kotlin to 1.4.30, turn on JS IR

CommanderTvis:


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Tue Dec 18 10:40:21 MSK 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.8-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.8.1-bin.zip
```
qurbonzoda, line 6: Are there any new capabiilities in Gradle 6.8.1 for `js(BOTH)` over 6.8?
If not, it would be better to leave the current Gradle version until the next library release.

qurbonzoda (commented):
Please rebase onto the master

gradle/wrapper/gradle-wrapper.properties
CommanderTvis: There aren't, I'll discard this change

plugin/gradle.properties
```diff
@@ -1,11 +1,14 @@
 kotlin.code.style=official
 
 # Set up your maven coordinates here, artifactId is defined per project.
-version = 0.3.0
+version = 0.2.0
```
qurbonzoda, line 4: Please reverse these changes

examples/kotlin-multiplatform/build.gradle
```diff
@@ -24,7 +24,7 @@ kotlin {
         }
     }
 
-    js {
+    js(BOTH) {
```
qurbonzoda, line 27: What does the `BOTH` mode supposed to mean for the non-library project?

runtime/build.gradle
```diff
@@ -21,7 +21,7 @@ kotlin {
         }
     }
 
-    js {
+    js(BOTH) {
```
qurbonzoda, line 24: While for the runtime library it is enough to publish atrifacts for jsIr, the plugin needs some adjustments to be able run benchmarks

qurbonzoda (changes requested):
While for the runtime library it is enough to publish artifacts for jsIr, the plugin needs some adjustments to be able run benchmarks

CommanderTvis:
> adjustments to be able run benchmarks
@qurbonzoda 

What should I do?

CommanderTvis:
@qurbonzoda Are there any integration tests or similar code to make all the needed adjustments? JS IR support is extremely important for my project.

whyoleg:
@CommanderTvis you can take a look on another PR, which contains setup for integration testing of plugin + runtime here #37. To support JS with both compilers it should be adapted (as you will need to put also klibs of JS IR compiler, similar to how it handled for native).

PR #32, opened 2021-02-15
Kotlin/Native support

LepilkinaElena:


build.gradle
```diff
@@ -2,6 +2,7 @@ buildscript {
     repositories {
         maven { url 'https://dl.bintray.com/kotlin/kotlin-dev' }
         jcenter()
+        mavenLocal()
```
StefMa, line 5: I guess this shouldn't commitet 😁

plugin/build.gradle
```diff
@@ -2,6 +2,7 @@ buildscript {
     repositories {
     	maven { url 'https://dl.bintray.com/kotlin/kotlin-dev' }
         mavenCentral()
+        mavenLocal()
```
StefMa, line 5: Same here.

runtime/build.gradle
```diff
@@ -2,6 +2,7 @@ buildscript {
     repositories {
     	maven { url 'https://dl.bintray.com/kotlin/kotlin-dev' }
         mavenCentral()
+        mavenLocal()
```
StefMa, line 5: Same here

runtime/build.gradle
```diff
@@ -13,6 +14,7 @@ plugins {
 repositories {
 	maven { url 'https://dl.bintray.com/kotlin/kotlin-dev' }
     mavenCentral()
+    mavenLocal()
```
StefMa, line 17: Same here

README.md
```diff
@@ -166,6 +162,8 @@ Available configuration options:
 * `iterationTimeUnit` – time unit for `iterationTime` (default is seconds)
 * `outputTimeUnit` – time unit for results output
 * `mode` – "thrpt" for measuring operations per time, or "avgt" for measuring time per operation
+* `iterationMode` – "external" for iterating in gradle in order to get correct Kotlin/Native runtime input in measurement,
+ "internal" can be used if it's known that measured code have no calls in K/N runtime that can influence on measurement unrepeatedly.
```
ilya-g, line 166: From this description it's very hard to get when one should use this configuration option. Perhaps, if there's a sensible default, we could turn it into a boolean option that would allow to override this default and explain why users would want to do it.

plugin/main/src/kotlinx/benchmark/gradle/NativeMultiplatformTasks.kt
```diff
@@ -116,30 +118,127 @@ fun Project.createNativeBenchmarkExecTask(
...
+
+        detailedConfigFiles.forEach { runConfig ->
+            val runConfigPath = runConfig.absolutePath
+            val lines = runConfig.readLines()
+            require(lines.size > 1) { "Wrong detailed configuration format" }
+            val currentConfigDescription = lines[1]
+
+            // Execute benchmark
+            if (config.iterationMode == "internal") {
+                val jsonFile = createTempFile("bench", ".json").toFile()
```
ilya-g, line 190: Just a suggestion: most `kotlin.io.*` functions have analogs in `kotlin.io.path`, so you can avoid converting the result of `createTempFile` to `File` and just use the returned `Path`.

README.md
LepilkinaElena: "External" is already such default. It's similar to other settings that also have. default values, but about description I agree. If you have ideas what to write. in README, I'll be greatful

plugin/main/src/kotlinx/benchmark/gradle/NativeMultiplatformTasks.kt
LepilkinaElena: The problem is that `kotlin.io.createTempFile` is deprecated

```
@Deprecated(
    "Avoid creating temporary files in the default temp location with this function " +
    "due to too wide permissions on the newly created file. " +
    "Use kotlin.io.path.createTempFile instead or resort to java.io.File.createTempFile."
)

I used one of suggested alternatives

File: README.md (hunk 5 above)
Comment (LepilkinaElena) [reply README.md:166]: `* iterationMode – way of iteration for K/N benchmarks. Default value - "external", each iteration is processed as separate run, allows to get more precise results in most cases. "internal" - all iterations of one benchmark run during one binary execution (pay attention, some code performance can't be measured properly using this mode, e.g. singleton initialization, because of implementation details of K/N runtime)`

Is this better?

File: plugin/main/src/kotlinx/benchmark/gradle/NativeMultiplatformTasks.kt [hunk 7]
```diff
@@ -116,30 +118,127 @@ fun Project.createNativeBenchmarkExecTask(
...
+            } else {
+                val iterations = currentConfigDescription.substringAfter("iterations=")
+                    .substringBefore(',').toInt()
+                val warmups = currentConfigDescription.substringAfter("warmups=")
+                    .substringBefore(',').toInt()
+                // Warm up
+                var exceptionDuringExecution = false
+                for(i in 0 until warmups) {
+                    val textResult = createTempFile("bench", ".txt").toFile()
+                    execute(listOf(configFile.absolutePath, "--warmup", runConfigPath, i.toString(), textResult.absolutePath))
```
qurbonzoda, line 202: What is the benefit of the warmups if they are executed in their own processes?

runtime/nativeMain/src/kotlinx/benchmark/native/NativeExecutor.kt
```diff
@@ -1,73 +1,172 @@
...
+    }
+
+    private fun List<BenchmarkDescriptor<Any?>>.getBenchmark(name: String) = find { it.name == name }
+        ?: throw NoSuchElementException("Benchmark $name wasn't found.")
+
+    private fun runBenchmarkIteration(benchmarks: List<BenchmarkDescriptor<Any?>>) {
+        val (configFileName, iteration, resultsFile) = additionalArguments
+        val benchmarkRun = configFileName.parseBenchmarkConfig()
+        val benchmark = benchmarks.getBenchmark(benchmarkRun.benchmarkName)
+        val samples = run(benchmark, benchmarkRun, iteration.toInt())
```
qurbonzoda, line 40: Does each measurment iteration run a warmup iteration first?

runtime/nativeMain/src/kotlinx/benchmark/native/NativeExecutor.kt
```diff
@@ -96,24 +197,30 @@ class NativeExecutor(name: String, args: Array<out String>) : SuiteExecutor(name
         name: String,
         config: BenchmarkConfiguration,
         instance: T,
-        benchmark: BenchmarkDescriptor<T>
+        benchmark: BenchmarkDescriptor<T>,
+        currentIteration: Int? = null
     ): Int {
         var iterations = 0
-        repeat(config.warmups) { iteration ->
+        val warmupIterations = if (config.iterationMode == IterationMode.External) 1 else config.warmups
```
qurbonzoda, line 204: Is one warmup iteration enough to get a reasonable measurement cycles number?
And is it enough for caches warmup?

plugin/main/src/kotlinx/benchmark/gradle/NativeMultiplatformTasks.kt
ilya-g: Sure, I mean you can use the result of `kotlin.io.path.createTempFile` without converting it `toFile()` first. Though I don't insist.

runtime/nativeMain/src/kotlinx/benchmark/Utils.kt
```diff
@@ -44,3 +54,18 @@ actual fun String.readConfigFile(): String = buildString {
...
+
+internal fun String.parseBenchmarkConfig(): NativeExecutor.BenchmarkRun {
+    fun String.getElement(name: String) =
+        if (startsWith(name)) {
+            substringAfter("$name: ")
+        } else throw NoSuchElementException("Parameter `$name` is required.")
+
+    val content = readFile()
+    val lines = content.lines().filter { it.isNotEmpty() }
+    require(lines.size == 3, { "Wrong fromat of detailed benchmark configuration file. "})
```
ilya-g, line 66:
```suggestion
    require(lines.size == 3, { "Wrong format of detailed benchmark configuration file. "})
```

README.md
ilya-g: It feels conceptually like 'fork' option in JMH (see the [example](https://github.com/openjdk/jmh/blob/master/jmh-samples/src/main/java/org/openjdk/jmh/samples/JMHSample_12_Forking.java)). Perhaps we could find a resembling name for this option.

ilya-g: > singleton initialization

I always thought that microbenchmarks tried to measure the settled code execution performance, i.e. when all fixed initialization costs are done and no longer affect the performance of the code being measured. That's why they do warm-up iterations.
So for me it looks like that the mode of a fork-per-benchmark method should be the preferred default.

LepilkinaElena: I thought during implementation about fork, but it's different in jmh. Firstly fork is number and user can set by user. Also as I undertood JMH created needed number of forks and warmups and iterations is made inside each fork, logic of working for "external" mode is another, forks are even more similar for "internal" mode where is one fork for each benchmark. I really think that using `fork` can confuse users, because they'll expect the same behaviour for Native.

LepilkinaElena: > when all fixed initialization costs are done and no longer affect the performance of the code being measured.

It isn't true for native runtime. Done initialization may influence on the next code. And we don't know what exactly user want to measure. 

> That's why they do warm-up iterations.

Warm-uo for native applications is needed to warmup processors caches, etc..

> So for me it looks like that the mode of a fork-per-benchmark method should be the preferred default.

Any code can influence on thresholds inside runtime and that can influence on performance, if user doesn't know runtime implementation details and want to get accurate results he should use "external". We can change default, but I amn't sure that it's right for majority of users. May be we should discuss this, it was already discussed with @qurbonzoda

plugin/main/src/kotlinx/benchmark/gradle/NativeMultiplatformTasks.kt
LepilkinaElena: May be I didn't get. What was your suggestion? Use deprecated function in order to eliminate `toFile()`?  Or do you mean using `Path`, not `File`?

plugin/main/src/kotlinx/benchmark/gradle/NativeMultiplatformTasks.kt
LepilkinaElena: Warm-up for native applications are needed to warmup processors caches, etc..

runtime/nativeMain/src/kotlinx/benchmark/native/NativeExecutor.kt
LepilkinaElena: Yes, because without extra single run we can't calculate iteration number in case if time is less then set iterationTime

runtime/nativeMain/src/kotlinx/benchmark/native/NativeExecutor.kt
LepilkinaElena: Warm up is already done, for cycles number it should be enough.

LepilkinaElena: May be logic of getting cycles number is worth to include inside iteration.

qurbonzoda: What about passing the number of cycles from actual warmup iterations? (e.g. through a temp file)
What bothers me is that a measurement iteration takes two times longer that it is supposed to take. It may be significant for long-running benchmarks, nightly runs

qurbonzoda: If warmup iteration time difers from measurement iterations time, the number could be adjusted by the measurement_time/warmup_time ratio. 
Alternatively, it is possible to run an initial iteration with measurement iterations time for calculating cycles without reporting the result of that iteration, and use the calculated cycles to run measurement iterations.

runtime/nativeMain/src/kotlinx/benchmark/NativeBlackhole.kt
```diff
@@ -1,13 +1,37 @@
...
-    actual fun consume(f: Float) {}
-    actual fun consume(d: Double) {}
+    @ThreadLocal
+    companion object {
+        var consumer = 0
+    }
+    actual fun consume(obj: Any?) {
+        // hashCode now is implemented as taking address of object, so it's suitable now.
+        // If implementation is changed `Blackhole` should be reimplemented.
+        consumer += obj.hashCode()
```
qwwdfsad, line 11: That's a virtual call that also may be way too heavy for a user-supplied class. That's just a time bomb waiting to explode, so I'd suggest using Blackhole approach (non-ordered volatile write, LLVM directive, w/e).


Also, see https://bugs.openjdk.java.net/browse/JDK-8252505 for the potential pitfalls

runtime/nativeMain/src/kotlinx/benchmark/NativeBlackhole.kt
```diff
@@ -1,13 +1,37 @@
...
+    }
+    actual fun consume(bool: Boolean) {
+        consumer += bool.hashCode()
```
qwwdfsad, line 14: Here and below, this is still a write, moreover, it is a write to `@ThreadLocal object` field, so it's unlikely to be `add` instruction to a known address.

Let's please avoid any writes in performance sensitive-code

runtime/nativeMain/src/kotlinx/benchmark/native/NativeExecutor.kt
```diff
@@ -82,11 +181,13 @@ class NativeExecutor(name: String, args: Array<out String>) : SuiteExecutor(name
     ): Double {
         val executeFunction = benchmark.function
         var counter = cycles
+        GC.collect()
         val startTime = getTimeNanos()
         while (counter-- > 0) {
             @Suppress("UNUSED_VARIABLE")
             val result = instance.executeFunction() // ignore result for now, but might need to consume it somehow
         }
+        GC.collect()
```
qwwdfsad, line 190: Unconditional GC between each measurement iteration implies multiple things:

- Small micro-nano benchmarks may be dominated by this runtime call with undesired side-effects
- The amortized cost of allocations (especially TLAB-like) is simply not measured. E.g. it would be impossible to compare two JSON parsers that differentiate mostly in allocation patterns 
- We potentially mess up with CPU caches on each iteration. That's not really helpful for correct measurements and comparison, especially when comparing regular and cache-obvious algorithms

I strongly suggest making GC after each iteration optional and disabled by default

qwwdfsad (changes requested):

runtime/nativeMain/src/kotlinx/benchmark/native/NativeExecutor.kt
LepilkinaElena: I got your concerns, I can do several modes. We can even use fully turned off GC as the second mode. I really don't like an idea to. run benchmark with only auto GC calls made by runtime, because then measurements can differ a lot between different runs of benchmarks. Moreover, some users expect that the result of running same benchmark as separate application should be close to results of benchmarking with library, that's impossible with auto-called GC.

What do you think if we make 2 modes: one with calling GC after each iteration (behaviour similar to separate application) and without GC calls at all?

qwwdfsad: IMO It's hard to actually have non-micro benchmarks without GC at all, but in my (JVM-ish) experience I mostly used to measure GC as amortized part of the execution. And in really rare cases (whether I want precision, a minimal amount of noise or do not care about memory pressure) I use `-prof gc` that collects garbage after each iteration.

I think it's ok to have three modes for GC: `none`, `auto` and `iteration` with `auto` as default

runtime/nativeMain/src/kotlinx/benchmark/native/NativeExecutor.kt
```diff
@@ -78,15 +204,19 @@ class NativeExecutor(name: String, args: Array<out String>) : SuiteExecutor(name
     private fun <T> measure(
         instance: T,
         benchmark: BenchmarkDescriptor<T>,
-        cycles: Int
+        cycles: Int,
+        nativeGCCollectMode: NativeGCCollectMode
     ): Double {
         val executeFunction = benchmark.function
         var counter = cycles
+        GC.collect()
```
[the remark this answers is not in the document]
qurbonzoda: warmup also does `GC.collect()`

qurbonzoda (changes requested):

runtime/nativeMain/src/kotlinx/benchmark/NativeBlackhole.kt
```diff
@@ -1,13 +1,178 @@
 package kotlinx.benchmark
 
+import kotlinx.cinterop.pin
+import kotlin.experimental.xor
```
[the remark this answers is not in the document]
LepilkinaElena: What comment do you want to see? That this is for another version of implementation?

plugin/main/src/kotlinx/benchmark/gradle/BenchmarkConfiguration.kt
```diff
@@ -9,6 +9,7 @@ open class BenchmarkConfiguration(val extension: BenchmarksExtension, val name:
     var iterationTime: Long? = null
     var iterationTimeUnit: String? = null
     var mode: String? = null
+    var nativeGCCollectMode: String? = null
```
qurbonzoda, line 12: I would rather make `nativeGCCollectMode` be declared via `advanced(...)` as well, similar to `nativeIterationMode`.

qurbonzoda (approved):

ilya-g:
Several discussions are left unresolved here, and I think the merge was too early. 
We should resolve outstanding questions with follow-up commits/PRs to the master.

LepilkinaElena:
It was discussed with @qurbonzoda in direct messages and it's decided to merge it, I got an approve to merge from him.

@ilya-g could you please write all your concerns.

PR #37, opened 2021-03-05
Integration testing

whyoleg:
* Implemented integration testing of gradle plugin + runtime for all platforms: JVM, JS, Native
* Simple POC test for testing of generated report file names

### DIFF
--- a/integration/build.gradle.kts
+++ b/integration/build.gradle.kts
@@ -1,0 +1,53 @@
+import org.jetbrains.kotlin.gradle.dsl.*
+
+plugins {
+    kotlin("jvm")
+}
+
+repositories {
+    mavenCentral()
+}
+
+evaluationDependsOn(":kotlinx-benchmark-runtime")
+
+val runtime get() = project(":kotlinx-benchmark-runtime")
+val plugin get() = gradle.includedBuild("plugin")
+
+val AbstractArchiveTask.archiveFilePath get() = archiveFile.get().asFile.path
+
+fun artifactsTask(artifact: String) = runtime.tasks.getByName<AbstractArchiveTask>("${artifact}Jar")
+fun artifactsTaskNativeKlibs() = runtime.tasks.getByName("compileKotlinNative")
+
+fun Task.klibs(): String = outputs.files.filter { it.extension == "klib" }.joinToString("\n")
+
+fun IncludedBuild.classpath() = projectDir.resolve("build/createClasspathManifest")
+
+val createClasspathManifest by tasks.registering {
+    dependsOn(plugin.task(":createClasspathManifest"))
+    dependsOn(artifactsTask("jvm"))
+    dependsOn(artifactsTask("js"))
+    dependsOn(artifactsTask("metadata"))
+    dependsOn(artifactsTask("nativeMetadata"))
+    dependsOn(artifactsTaskNativeKlibs())
+
+    val outputDir = file("$buildDir/$name")
+    outputs.dir(outputDir)
+    doLast {
+        outputDir.apply {
+            mkdirs()
+            resolve("plugin-classpath.txt").writeText(plugin.classpath().resolve("plugin-classpath.txt").readText())
+            resolve("runtime-metadata.txt").writeText(artifactsTask("metadata").archiveFilePath)
+            resolve("runtime-jvm.txt").writeText(artifactsTask("jvm").archiveFilePath)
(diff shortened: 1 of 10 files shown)


integration/build.gradle.kts
```diff
@@ -0,0 +1,53 @@
...
+
+evaluationDependsOn(":kotlinx-benchmark-runtime")
+
+val runtime get() = project(":kotlinx-benchmark-runtime")
+val plugin get() = gradle.includedBuild("plugin")
+
+val AbstractArchiveTask.archiveFilePath get() = archiveFile.get().asFile.path
+
+fun artifactsTask(artifact: String) = runtime.tasks.getByName<AbstractArchiveTask>("${artifact}Jar")
+fun artifactsTaskNativeKlibs() = runtime.tasks.getByName("compileKotlinNative")
```
qurbonzoda, line 19: The Gradle task `compileKotlinNative` only exists when it is run in the IDEA environment because of the `kotlinx.team.infra` plugin, which merges the same tasks of all native targets under "native" wording.
That is `compileKotlinLinuxX64`, `compileKotlinMingwX64`, and `compileKotlinMacosX64` are unified to `compileKotlinNative`, and running that task delegates to the host task.

However, Gradle won't find `compileKotlinNative` if you run it from the terminal. You can workaround this with smth  like:
```
val nativeHost = if (System.getProperty("idea.active") == "true") {
    "native"
} else when {
    org.jetbrains.kotlin.konan.target.HostManager.hostIsLinux -> "linuxX64"
    org.jetbrains.kotlin.konan.target.HostManager.hostIsMingw -> "mingwX64"
    org.jetbrains.kotlin.konan.target.HostManager.hostIsMac -> "macosX64"
    else -> error("Unknown host: ${org.jetbrains.kotlin.konan.target.HostManager.host}")
}
```

qurbonzoda: Please rebase on master first and then use `getSystemProperty("idea.active")` instead of `System.getProperty`.
Here is the reason why: https://github.com/Kotlin/kotlinx-benchmark/pull/40

integration/src/main/kotlin/kotlinx/benchmark/integration/BenchmarkConfiguration.kt
```diff
@@ -0,0 +1,25 @@
+package kotlinx.benchmark.integration
+
+class BenchmarkConfiguration {
```
qurbonzoda, line 3: Is it possible to reuse the existing class `kotlinx.benchmark.gradle.BenchmarkConfiguration`?

integration/build.gradle.kts
whyoleg: Ok, will do it
I was thinking about it, that it could not work on CI:)

integration/src/main/kotlin/kotlinx/benchmark/integration/BenchmarkConfiguration.kt
whyoleg: Hm, will check it, but anyway, we need to put those values into `build.gradle` inside temp project used for testing.

integration/build.gradle.kts
whyoleg: it's internal in plugin + we don't have plugin at classpath there, I was just copied this functions. Not sure it's best way, maybe you can advise something else

integration/src/main/kotlin/kotlinx/benchmark/integration/BenchmarkConfiguration.kt
whyoleg: On current moment, plugin is not in classpath of integration module, because `plugin` is included build (to use current version of plugin in examples).
We can add it to classpath in some way, but now `kotlinx.benchmark.gradle.BenchmarkConfiguration` depends on `BenchmarksExtension`, looks like we need to remove this dependency to reuse it.

I have a question, why is included build is used at all? Maybe it's possible to make it as an ordinary module, and after that integration module will be moved just to `integrationTest` source set under it (the same way as it's done in `kotlinx-atomicfu` and `binary-compatibility-validator` gradle plugins)

qurbonzoda: > why is included build is used at all?

Seems like for historical reasons. Let's explore the possibility of making the plugin an ordinal module in a separate effort.

PR #40, opened 2021-04-27
Declare system properties reads to be compatible with Gradle configuration cache

ALikhachev:
Gradle reports direct system properties reads as configuration cache problems as system properties are potential build configuration inputs.
See https://docs.gradle.org/7.0/userguide/configuration_cache.html for more details


qurbonzoda:
Couldn't reproduce the warning or problem with enabled configuration cache, despite the fact that gradle doc defines direct system properties reads as problems.
Were you able to get problem report related to those specific system property reads?

ALikhachev:
Gradle reports system property read as a problem if it's value is not `null` for easier adoption. (https://github.com/gradle/gradle/blob/12254f28d6d9b35f7e352517716d3a1ab8d0989a/subprojects/configuration-cache/src/main/kotlin/org/gradle/configurationcache/SystemPropertyAccessListener.kt#L81)
You can try to reproduce it by manually setting some value for these system properties or by running some task from IDEA

qurbonzoda:
Thank you!

ilya-g:
@ALikhachev we've released the version 0.3.1 with this fix.

PR #41, opened 2021-05-03
Exclude duplicate classes for Jar task

### LINKED ISSUE #39: Wrong treatment of dublication META-INF elements
The jvmBenchmarkJar task is failing with complex dependencies on Gradle 7 unless the following fix is applied:
```kotlin
    val jvmBenchmarkJar by tasks.getting(org.gradle.jvm.tasks.Jar::class) {
        duplicatesStrategy = org.gradle.api.file.DuplicatesStrategy.EXCLUDE
    }
```

fabianishere:
Hi,

This pull request attempts to fix issue #39, which causes a failure for benchmarks with complex dependencies when using Gradle 7. This issue is caused by the default `duplicatesStrategy` of the `Jar` task triggering a failure due to complex dependencies (where the same class might exist multiple times).
 
The solution is to update the `duplicatesStrategy` for the `jmhBenchmarkJar` tasks to exclude duplicate classes.

Fixes #39


qurbonzoda:
Thanks for your contribution!

PR #38, opened 2021-03-28
Support JS IR in plugin

whyoleg:
* fixes #30 
* suppresses #31
* TODO `ms` time units fails on IR

PR #49, opened 2021-09-07
Upgrade kotlin to 1.5.30

qurbonzoda:


qurbonzoda:
Kotlin 1.5.30 will also facilitate JS-IR support: https://github.com/Kotlin/kotlinx-benchmark/pull/38

README.md
```diff
@@ -25,7 +25,7 @@ the hoods to run benchmarks on JVM.
 
 Gradle 6.8 or newer
```
ilya-g, line 26: Does the minimal supported Gradle version have Kotlin 1.5 bundled? If not, it may be not safe to use Kotlin stdlib 1.5 API in the code of the Gradle plugin.

Issue #52, opened 2021-09-27
Implement easy publishing to a custom Maven repository


SvyatoslavScherbina:
To adopt kotlinx.benchmark in Kotlin/Native own performance infrastructure, we need to be able to easily publish the library and the plugin to a custom Maven repository (probably somewhere on https://maven.pkg.jetbrains.space) from a branch.

Motivation: we'd like to have a fast feedback loop, because migrating to kotlinx.benchmark would probably require some adjustments to the latter. Same for the further maintenance.

qurbonzoda:
Run of `Deploy(Publish)` build configuration now prompts you to select between "space" and "sonatype" repositories to publish artifacts to. Once space is selected and artifacts are published, they become available in https://maven.pkg.jetbrains.space/kotlin/p/kotlinx/dev repository.

PR #59, opened 2021-10-18
Publish :plugin and :runtime to Space besides GPP and Sonatype

qurbonzoda:


qurbonzoda:
Addresses https://github.com/Kotlin/kotlinx-benchmark/issues/52

PR #61, opened 2021-10-25
Improve support for quotation marks in params

goncalossilva:
This PR adds support for parameters that contain quotation marks, such as `@Param("""a "string" with quotes""")` by improving code generation and report formatting.

### DIFF
--- a/plugin/main/src/kotlinx/benchmark/gradle/SuiteSourceGenerator.kt
+++ b/plugin/main/src/kotlinx/benchmark/gradle/SuiteSourceGenerator.kt
@@ -208,15 +208,15 @@ class SuiteSourceGenerator(val title: String, val module: ModuleDescriptor, val 
                     val annotation = it.annotations.findAnnotation(FqName(paramAnnotationFQN))!!
                     val constant = annotation.argumentValue("value") 
                         ?: error("@Param annotation should have at least one default value")
-                    val values = constant.value as? List<String>
+                    val values = constant.value as? List<StringValue>
                         ?: error("@Param annotation should have at least one default value")
                     values
                 })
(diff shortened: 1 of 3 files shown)


plugin/main/src/kotlinx/benchmark/gradle/SuiteSourceGenerator.kt
```diff
@@ -208,15 +208,15 @@ class SuiteSourceGenerator(val title: String, val module: ModuleDescriptor, val
...
                         ?: error("@Param annotation should have at least one default value")
                     values
                 })
                 
                 val defaultParametersString = defaultParameters.entries
-                    .joinToString(prefix = "mapOf(", postfix = ")") {  
-                        "\"${it.key}\" to " + it.value.joinToString(prefix = "listOf(", postfix = ")")
-                        
+                    .joinToString(prefix = "mapOf(", postfix = ")") { (key, value) ->
+                        "\"${key}\" to ${value.joinToString(prefix = "listOf(", postfix = ")") { "\"${it.value.replace("\"", "\\\"")}\"" }}"
```
qurbonzoda, line 218: What about escaping other special characters?

qurbonzoda:
Thank you!

plugin/main/src/kotlinx/benchmark/gradle/SuiteSourceGenerator.kt
qurbonzoda: I have fixed it with the following commit: https://github.com/Kotlin/kotlinx-benchmark/commit/85ddb0655275e54a4ee62760dbed29db2f548650

PR #64, opened 2021-11-24
Support JS IR

### LINKED ISSUE #30: Turn on JS IR
I'm having such an error because the runtime library doesn't support JS IR.

```
Caused by: org.gradle.internal.component.NoMatchingConfigurationSelectionException: No matching variant of org.jetbrains.kotlinx:kotlinx.benchmark.runtime:0.2.0-dev-20 was found. The consumer was configured to find a usage of 'kotlin-runtime' of a component, as well as attribute 'org.jetbrains.kotlin.platform.type' with value 'js', attribute 'org.jetbrains.kotlin.js.compiler' with value 'ir' but:
  - Variant 'js-api' capability org.jetbrains.kotlinx:kotlinx.benchmark.runtime:0.2.0-dev-20 declares a component, as well as attribute 'org.jetbrains.kotlin.platform.type' with value 'js':
 ```

### PR DESCRIPTION
**Pull request by olme04**
* fixes #30 (actualized #38)
* share klib resolving logic between JS IR and Native
* minor cleanup of buildscripts
* add support for JS IR in integration tests
* add support for JS IR in example

_Note: user should always specify, if they want to use IR or Legacy for running benchmarks. It will not work when specified BOTH, as target with BOTH can not have `executables`_

### DIFF
--- a/integration/src/test/resources/templates/kotlin-multiplatform/build.gradle
+++ b/integration/src/test/resources/templates/kotlin-multiplatform/build.gradle
@@ -2,15 +2,9 @@ import org.jetbrains.kotlin.konan.target.KonanTarget
 import org.jetbrains.kotlin.konan.target.HostManager
 
 kotlin {
-    jvm {
-        compilations.all {
-            kotlinOptions.jvmTarget = '1.8'
-        }
-    }
-
-    js {
-        nodejs()
-    }
+    jvm()
+    js('jsIr', IR) { nodejs() }
+    js('jsLegacy', LEGACY) { nodejs() }
 
     if (HostManager.hostIsLinux) linuxX64('native')
     if (HostManager.hostIsMingw) mingwX64('native')
@@ -28,9 +22,14 @@ kotlin {
                 implementation(benchmarkRuntimeJvm)
             }
         }
-        jsMain {
+        jsIrMain {
             dependencies {
-                implementation(benchmarkRuntimeJs)
+                implementation(benchmarkRuntimeJsIr)
+            }
+        }
+        jsLegacyMain {
+            dependencies {
+                implementation(benchmarkRuntimeJsLegacy)
             }
         }
         nativeMain {
@@ -45,7 +44,8 @@ kotlin {
 benchmark {
     targets {
         register("jvm")
-        register("js")
+        register("jsIr")
+        register("jsLegacy")
         register("native")
     }
 }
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

**Comment by olme04** [comment]
any news on this ? :)

**Comment by qurbonzoda** [comment]
Thank you for your contribution!
Sorry for the late review.

**Comment by olme04** [comment]
Cool, thx!
When can we expect it will be released?

**Comment by qurbonzoda** [comment]
Approximately after a week, when vacation ends.

File: integration/src/test/resources/templates/kotlin-multiplatform/build.gradle [hunk 1]
```diff
@@ -2,15 +2,9 @@ import org.jetbrains.kotlin.konan.target.KonanTarget
...
-            kotlinOptions.jvmTarget = '1.8'
-        }
-    }
-
-    js {
-        nodejs()
-    }
+    jvm()
+    js('jsIr', IR) { nodejs() }
+    js('jsLegacy', LEGACY) { nodejs() }
```
max-kammerer, line 7: Why not simple `js {nodejs()}`. `jsLegacy` could be confusing as new backend still not stable and not default. It's also inconsistent with kotlin gradle plugin dsl, where `js  {}` is used

olme04: it's template for integration tests, to test for both LEGACY and IR
In example, yes, this could be changed to `js`.
And then, somewhere should be noted, that `IR` is also supported. 
And if users will use `kotlin.js.compiler=both` or `js(BOTH)` - this will not work, as `BOTH` doesn't support generating  executables

max-kammerer: @qurbonzoda @anton-bannykh what do you think?

qurbonzoda: Indeed `js('jsLegacy', LEGACY)` could be just `js {}`, but I think for the test project used in integration tests it's not that critical. It makes sense to fix it in `kotlin-multiplatform` example project though.

I will update README with the IR support information when it gets released.

qurbonzoda: `jsLagacy` is now replaced to `js`: https://github.com/Kotlin/kotlinx-benchmark/pull/77

PR #73, opened 2022-01-04
Fix `mode` config to consist with docs

guohao:
Error occurred when set `mode` to `avgt` according to docs.
```kotlin
benchmark {
    configurations {
        named("main") {
            mode="avgt"
        }
    }
```

> Exception in thread "main" java.lang.IllegalArgumentException: No enum constant org.openjdk.jmh.annotations.Mode.avgt

But set `mode='AverageTime'` works.

In order to consist with docs, I modify the `ConfigurationReader`


qurbonzoda:
Thank you!

PR #107, opened 2023-02-10
remove wrong whitespace in readme

jisungbin:
Thanks for creating the awesome repository. I found the wrong whitespace, so removed it.

` benchmark` -> `benchmark`


qurbonzoda:
Thank you for the improvement!

PR #108, opened 2023-02-13
Fix for building large benchmark JARs

### LINKED ISSUE #95: Add support for zip64 or shadowJar option
For complex project with multiple dependencies, kotlinx-benchmark fails to create JMH executable JAR with following error:
```
Execution failed for task ':benchmark:mainBenchmarkJar'.
> archive contains more than 65535 entries.

```
A quick google search revealed that for shadow or uber JAR generation, using zip64 alleviates this issue.

Yeicor:
This should be enough to support building large JARs.

It prevents this error when building JARs with more than 65535 entries or larger than 4GB:

```
Execution failed for task ':mainBenchmarkJar'.
> archive contains more than 65535 entries.
```

Note that it may affect the compatibility of the JAR files with some old tools (see https://docs.gradle.org/current/dsl/org.gradle.api.tasks.bundling.Jar.html#org.gradle.api.tasks.bundling.Jar:zip64)

Fixes #95

### DIFF
--- a/plugin/main/src/kotlinx/benchmark/gradle/JvmTasks.kt
+++ b/plugin/main/src/kotlinx/benchmark/gradle/JvmTasks.kt
@@ -28,6 +28,7 @@ fun Project.createJvmBenchmarkCompileTask(target: JvmBenchmarkTarget, compileCla
     ) {
         group = BenchmarksPlugin.BENCHMARKS_TASK_GROUP
         description = "Build JAR for JMH compiled files for '${target.name}'"
+        isZip64 = true
         dependsOn("${target.name}${BenchmarksPlugin.BENCHMARK_COMPILE_SUFFIX}")
         archiveClassifier.set("JMH")
         manifest.attributes["Main-Class"] = "org.openjdk.jmh.Main"


qurbonzoda:
Hi Yeicor,
Thanks for the contribution!

I think JAR generation should be the default behavior. And an option should be provided in the Gradle plugin to enable zip64. 
What do you think about such a setup?

Yeicor:
Yes, that would be a better solution.

Maybe it could also detect the error automatically by counting (number of files and sizes) while listing the files and notify the user of the solution or fix it.

ilya-g:
I believe we need to maintain compatibility only with those tools that intend to read this jar. Therefore `zip64` could be used by default for building benchmarks jar.

Yeicor:
I just checked the implementation of setZip64 and it does not actually force to use the zip64 version unless it is actually needed.

Hence, this patch will:
- produce the same JARs that could be built previously, without reducing the compatibility with old tools.
- allow building large JARs that previously failed, automatically switching to zip64 mode when required.

ilya-g (approved):

Issue #96, opened 2022-07-31
Examples don't work for Kotlin Multiplatform project


aSemy:
I've tried copying the examples 

* https://github.com/Kotlin/kotlinx-benchmark/blob/24fb001e8cae88befd5b352b0e281517fe595f67/examples/kotlin-multiplatform/build.gradle
* https://github.com/Kotlin/kotlinx-benchmark/blob/24fb001e8cae88befd5b352b0e281517fe595f67/examples/kotlin-kts/build.gradle.kts

but I can't make any good progress, I immediately come across fatal errors.

```
Cannot find a benchmark compilation 'main', ignoring.
```

```
* Exception is:
org.gradle.api.ProjectConfigurationException: A problem occurred configuring project ':modules:client-tests'

Caused by: java.lang.IllegalStateException: Could not create domain object 'main' (BenchmarkTarget)

Caused by: java.lang.ClassCastException: class kotlinx.benchmark.gradle.BenchmarkTarget cannot be cast to class kotlinx.benchmark.gradle.JvmBenchmarkTarget (kotlinx.benchmark.gradle.BenchmarkTarget and kotlinx.benchmark.gradle.JvmBenchmarkTarget are in unnamed module of loader org.gradle.internal.classloader.VisitableURLClassLoader @6e21e3fb)
```

* Kotlin 1.7.10
* Kotlinx Benchmark 0.4.4

```kotlin
import kotlinx.benchmark.gradle.JvmBenchmarkTarget

plugins {
    buildsrc.convention.`kotlin-multiplatform` // convention plugin that applies kotlin("multiplatform") 1.7.10

    id("org.jetbrains.kotlinx.benchmark")
    kotlin("plugin.allopen")
}

kotlin {
    jvm()

    sourceSets {
        val commonMain by getting {
            dependencies {
                implementation(kotlin("reflect"))

                implementation("org.jetbrains.kotlinx:kotlinx-benchmark-runtime:0.4.4")
            }
        }

        val commonTest by getting {
            dependencies {
                implementation(projects.modules.mockk)
            }
        }

        val jvmMain by getting {
            dependencies {
            }
        }

        val jvmTest by getting {
            dependencies {
                implementation(buildsrc.config.Deps.Libs.kotlinTestJunit()) {
                    exclude(group = "junit", module = "junit")
                }

                implementation(buildsrc.config.Deps.Libs.slfj)
                implementation(buildsrc.config.Deps.Libs.logback)

[...]

PR #119, opened 2023-06-27
docs: add READMEs to examples

wldeh:
Addresses #96 

This pull request introduces README files to each example subdirectory and the root examples directory. These README files serve as a guide to help both new and experienced users navigate and understand the structure and use of each example. The goal here is to improve the user experience by providing helpful and clear information about each example, making it easier for developers to get started with or dive deeper into the kotlinx-benchmark library.

Please review and let me know if there are any suggestions or improvements. Thanks!


examples/README.md
```diff
@@ -0,0 +1,59 @@
...
+1. **Navigate to the Example Directory**: Navigate to the directory of the example you want to run. For instance, if you want to run the `kotlin-kts` example, you would run:
+
+   ```
+   cd kotlin-kts
+   ```
+
+2. **Build the Project**: Each project uses Gradle as a build tool. If you have Gradle installed on your machine, you can build the project by running:
+
+   ```
+   gradle build
```
qurbonzoda, line 39: Using Wrapper is preferred. See for more information: https://docs.gradle.org/current/userguide/gradle_wrapper.html
https://stackoverflow.com/questions/39627231/difference-between-using-gradlew-and-gradle

`gradle` command uses locally installed gradle, which may vary user by user.
For instance, on my local machine `gradle build` in kotlin-kts directory fails.

All tasks in example projects should be run from the library root directory, where the wrapper is generated.
e.g.,  `./gradlew :examples:kotlin-kts:build`

examples/README.md
```diff
@@ -0,0 +1,59 @@
...
+   ```
+
+3. **Run the Benchmark**: After the project is built, you can run the benchmark by executing:
```
qurbonzoda, line 42: Is building the project required before running benchmarks?
I can run `./gradlew :examples:kotlin-kts:benchmark` without building the kotlin-kts project.

examples/java/README.md
```diff
@@ -0,0 +1,36 @@
...
+└── src/
+    └── main/
+        └── java/
+            └── test/
+                └── SampleJavaBenchmark.java
+```
+
+## Tasks
+
+All tasks can be run from the root of the project, from a terminal:
```
qurbonzoda, line 21: The tasks should be run from the root of the library

examples/java/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+| Task Name | Action |
+| --- | --- |
+| `gradle assembleBenchmarks` | Generate and build all benchmarks in the project |
```
qurbonzoda, line 25: `gradle` command should be removed in the Task name column.
WDYT?

examples/java/README.md
```diff
@@ -0,0 +1,36 @@
...
+| `gradle benchmark` | Execute all benchmarks in the project |
+| `gradle mainBenchmark` | Execute benchmark for 'main' |
+| `gradle mainBenchmarkCompile` | Compile JMH source files for 'main' |
+| `gradle mainBenchmarkGenerate` | Generate JMH source files for 'main' |
+| `gradle mainBenchmarkJar` | Build JAR for JMH compiled files for 'main' |
+| `gradle mainSingleParamBenchmark` | Execute benchmark for 'main' |
+| `gradle singleParamBenchmark` | Execute all benchmarks in the project |
```
qurbonzoda, line 32: Could you please specify what is the different between `singleParamBenchmark` and `benchmark` tasks?
They have exactly the same description. Their description should mention the configuration name the task executes benchmarks from.

examples/java/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Tasks
+
+All tasks can be run from the root of the project, from a terminal:
+
+| Task Name | Action |
+| --- | --- |
+| `gradle assembleBenchmarks` | Generate and build all benchmarks in the project |
+| `gradle benchmark` | Execute all benchmarks in the project |
+| `gradle mainBenchmark` | Execute benchmark for 'main' |
```
qurbonzoda, line 27: Could you please specify what 'main' stands for in the task description. It can be done in `createJvmBenchmarkExecTask` function in JvmTasks.kt. Basically, 'main' is the name of the source set registered as a benchmark target.

examples/kotlin-kts/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Tasks
+
+All tasks can be run from the root of the project, from a terminal:
+
+| Task Name | Action |
+| --- | --- |
+| `gradle assembleBenchmarks` | Generate and build all benchmarks in the project |
+| `gradle benchmark` | Execute all benchmarks in the project |
+| `gradle mainBenchmark` | Execute benchmark for 'benchmarks' |
```
qurbonzoda, line 25: Should be 'main' in this example project.

qurbonzoda (changes requested):
It seems a great opportunity to improve description of the tasks the plugin creates.

examples/README.md
wldeh: It is not. Understood and updated in the latest commit.

examples/README.md
wldeh: Noted. I have updated all the docs to utilize the wrapper `./gradlew`

examples/java/README.md
wldeh: I personally think it could be great to just copy and paste for folks using the terminal however I understand your point. ./gradlew does look a little bit messy however so I have omitted it as recommended.

examples/java/README.md
wldeh: Thank you for pointing this out. I have addressed this in the latest commit and added better descriptions to every task.

examples/kotlin-kts/README.md
wldeh: Thank you for your pointing this out and for your feedback.

qurbonzoda: This comment doesn't seem to be resolved.
Should be something like `Runs benchmarks for 'main' benchmark target`.

wldeh: My apologies, I believe I resolved it and lost my changes without saving. Will address

Issue #75, opened 2022-01-12
Align default configurations of a benchmark method on all targets


qurbonzoda:
Currently, if warmup (or measurement) iteration count is not defined in the configuration script and in the annotation for a benchmark, the default value in JVM is `5` while in JS and Native it is `3`. Default warmup (or measurement) iteration time in JVM is `10` seconds while in JS and Native it is `1` second.

Find out differences in other default values and consider aligning them.

qurbonzoda:
Will be released in 0.4.9

PR #120, opened 2023-06-28
Align default benchmark configurations

wldeh:
Addresses issue #75 by aligning the default warmup iteration time and count. The default number of warmup and measurement iterations is set to 5, and the default iteration time is set to 10 seconds, aligning with the JVM defaults.

### DIFF
--- a/runtime/commonMain/src/kotlinx/benchmark/SuiteDescriptor.kt
+++ b/runtime/commonMain/src/kotlinx/benchmark/SuiteDescriptor.kt
@@ -1,9 +1,11 @@
 package kotlinx.benchmark
 
 object DefaultDescriptorParameters {
-    val iterations = 3
-    val warmups = 3
-    val iterationTime = IterationTime(1, BenchmarkTimeUnit.SECONDS)
+    val iterations = 5
+    val warmups = 5
+    val iterationTime = IterationTime(10, BenchmarkTimeUnit.SECONDS)
+    val outputTimeUnit = BenchmarkTimeUnit.SECONDS
+    val mode = Mode.Throughput
 }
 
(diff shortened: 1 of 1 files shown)


runtime/commonMain/src/kotlinx/benchmark/SuiteDescriptor.kt
```diff
@@ -1,9 +1,9 @@
 package kotlinx.benchmark
 
 object DefaultDescriptorParameters {
```
qurbonzoda, line 3: Could you please move `SuiteDescriptor.outputTimeUnit` and `SuiteDescriptor.mode` default values to this object as well?
Also please make sure the specified defaults are the same in JVM.

qurbonzoda (changes requested):

runtime/commonMain/src/kotlinx/benchmark/SuiteDescriptor.kt
wldeh: Absolutely, pushing an update here shortly addressing your concerns.

We might want to consider decoupling the JVM default values from JMH as well, since the JMH defaults are subject to change (though unlikely). https://github.com/Kotlin/kotlinx-benchmark/blob/master/runtime/jvmMain/src/kotlinx/benchmark/jvm/JvmBenchmarkRunner.kt#L17

wldeh: For the time being, I have aligned them with the JMH defaults https://github.com/openjdk/jmh/blob/master/jmh-core/src/main/java/org/openjdk/jmh/runner/Defaults.java.

PR #123, opened 2023-06-29
Simplify integration tests build script

qurbonzoda:


integration/build.gradle.kts
```diff
@@ -10,75 +8,17 @@ repositories {
...
-
 dependencies {
-    implementation(files(createClasspathManifest))
     implementation(gradleTestKit())
 
     testImplementation(kotlin("test-junit"))
 }
 
 tasks.test {
+    dependsOn(plugin.task(":publishToMavenLocal"))
```
whyoleg, line 21: may be it's better not to use `mavenLocal` here, but create separate local repository inside project `buildDir`?
To not accidentally fetch other dependencies from `mavenLocal` + be able to easily cleanup it?

integration/src/main/kotlin/kotlinx/benchmark/integration/ProjectBuilder.kt
```diff
@@ -34,8 +32,13 @@ benchmark {
 private val buildScript = run {
     """
     buildscript {
+        repositories {
+            mavenCentral()
+            maven { url '${System.getProperty("plugin_repo_url")}' }
+        }
         dependencies {
-            classpath files(${readFileList("plugin-classpath.txt")})
+            classpath 'org.jetbrains.kotlin:kotlin-gradle-plugin:1.8.21'
```
qurbonzoda, line 40: The Kotlin version should be passed as a system property here as well. Because integration tests are executed in aggregate builds as well.

integration/build.gradle.kts
```diff
@@ -10,75 +8,21 @@ repositories {
...
-
 dependencies {
-    implementation(files(createClasspathManifest))
     implementation(gradleTestKit())
 
     testImplementation(kotlin("test-junit"))
 }
 
 tasks.test {
+    dependsOn(plugin.task(":publishToBuildLocal"))
```
whyoleg, line 21: where does this task come from? :) from `kotlinx-team-infra` plugin?

whyoleg (commented):
LGTM

integration/build.gradle.kts
qurbonzoda: Yeap, from kotlinx-team-infra plugin.

examples/java/build.gradle
```diff
@@ -1,6 +1,6 @@
 buildscript {
     dependencies {
-        classpath "org.jetbrains.kotlinx:kotlinx-benchmark-plugin:0.4.8"
+        classpath "org.jetbrains.kotlinx:kotlinx-benchmark-plugin:0.4.255-SNAPSHOT"
```
ilya-g, line 3: 255 is rather arbitrary here. In the main Kotlin repository we use this value because it's the maximum value a Kotlin component version can have, but here we have no such limitation.
Maybe it would be better to use an approach of other libraries, i.e. as soon as we released a version 0.4.x, we advance the version to 0.4.(x+1)-SNAPSHOT? Or, to avoid changing the version in examples after an each release, immediately to 0.5.0-SNAPSHOT?

.teamcity/settings.kts
```diff
@@ -72,7 +72,7 @@ fun Project.buildVersion() = BuildType {
     commonConfigure()
 
     params {
-        param(versionSuffixParameter, "SNAPSHOT")
+        param(versionSuffixParameter, "")
```
ilya-g, line 75: Is it possible to have this preserved when the settings are regenerated after the next kotlinx.team.infra plugin update?

examples/java/build.gradle
ilya-g, line 3: Do I get right that it works on CI due to the tests are never being run with an overridden version parameter?

integration/src/main/kotlin/kotlinx/benchmark/integration/ProjectBuilder.kt
```diff
@@ -31,35 +29,31 @@ benchmark {
...
+    else
+        "maven { url '$it' }"
+}
+
+private val buildScript =
     """
     buildscript {
+        repositories {
+            $kotlin_repo
+            maven { url '${System.getProperty("plugin_repo_url")}' }
```
ilya-g, line 44: Do we get all required transitive dependencies if we override `kotlin_repo` to something other than `mavenCentral`?

plugin/main/src/kotlinx/benchmark/gradle/BenchmarksPlugin.kt
```diff
@@ -7,7 +7,7 @@ import org.gradle.util.*
 class BenchmarksPlugin : Plugin<Project> {
     companion object {
         const val PLUGIN_ID = "org.jetbrains.kotlinx.benchmark"
-        const val PLUGIN_VERSION = "0.4.8"
+        const val PLUGIN_VERSION = "0.4.255-SNAPSHOT"
```
ilya-g, line 10: Is this value correctly overridden when we're building a build for deployment?

qurbonzoda: Now there is a task that overrides the value during a release build.

integration/src/main/kotlin/kotlinx/benchmark/integration/ProjectBuilder.kt
qurbonzoda: Now the mavenCentral repo is always in repositories list.

examples/java/build.gradle
qurbonzoda: I guess it's because the plugin is an included build here, so the specified version seems to be ignored.

.teamcity/settings.kts
qurbonzoda: This change is now moved to `additionalConfiguration.kt`. Thanks!

examples/java/build.gradle
qurbonzoda: `0.5.0-SNAPSHOT` is now used. Release version is requested when a TeamCity release build is run.

ilya-g (approved):

PR #127, opened 2023-07-05
Fix parsing of the Measurement annotation in SuiteSourceGenerator

qurbonzoda:


plugin/main/src/kotlinx/benchmark/gradle/SuiteSourceGenerator.kt
```diff
@@ -143,7 +143,7 @@ class SuiteSourceGenerator(val title: String, val module: ModuleDescriptor, val
 
         val measureIterations = measureAnnotation?.argumentValue("iterations")?.value as? Int
         val measureIterationTime = measureAnnotation?.argumentValue("time")?.value as? Int
-        val measureIterationTimeUnit = measureAnnotation?.argumentValue("timeUnit")?.value as? EnumValue
```
qurbonzoda, line 146: The error caused the `val` always be `null` as if `timeUnit` wasn't provided. Thus the default "SECONDS" was always assumed.

ilya-g: If it is expected to get either EnumValue | null here, maybe make the check more strict, i.e. `as EnumValue?`

qurbonzoda: Done

ilya-g (approved):

PR #129, opened 2023-07-10
[wasm] fix name of wasm target onto wasmJs

ilgonmic:


qurbonzoda:
Could you please clarify why this change is happening?

ilgonmic:
> Do we need to apply such changes to all libraries?

Now only benchmark is failing CI, so looks like other libraries can be migrated later

PR #131, opened 2023-07-21
Improve readme

wldeh:
Introduce a writing-benchmarks.md and polish README.


docs/writing-benchmarks.md
```diff
@@ -0,0 +1,77 @@
...
+
+#### @Setup
+
+The `@Setup` annotation is used to mark a method that sets up the necessary preconditions for your benchmark test. This method, annotated with `@Setup`, is executed before the actual benchmark methods are run. It serves as a preparatory step where you set up the environment for the benchmark, performing tasks such as generating data, establishing database connections, or preparing any other resources your benchmark requires. The key point to remember is that the `@Setup` method's execution time is not included in the final benchmark results - the timer starts only when the `@Benchmark` method begins. This makes `@Setup` an ideal place for initialization tasks that should not impact the timing results of your benchmark. By using the `@Setup` annotation, you ensure consistency across all executions of your benchmark, providing accurate and reliable results. In the provided example, the `@Setup` annotation is used to prepare a list of integers from 0 to a specified parameter value before the benchmark method is run.
```
qurbonzoda, line 45: Are `@Setup` and `@State` annotations related in any way?

qurbonzoda, line 45: Maybe "prepare a list **containing** integers from 0 to a specified parameter value" ?

qurbonzoda, line 45: What levels can the `Setup` annotation have?

docs/writing-benchmarks.md
```diff
@@ -0,0 +1,77 @@
...
+
+#### @Benchmark
+
+The `@Benchmark` annotation is used to specify the methods that you want to measure the performance of. Basically, it's the actual test you're running. The code you want to benchmark goes inside this method. In our example, the `benchmarkMethod` function is annotated with `@Benchmark`, which means the toolkit will measure the performance of the operation of summing all the integers in the list.
```
qurbonzoda, line 49: Are there any constrains related to the visibility modifier of the benchmark method?

docs/writing-benchmarks.md
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @Benchmark
+    fun benchmarkMethod(): Int {
+        return list.sum()
+    }
+
+    @TearDown
+    fun cleanup() {
+        list = ArrayList()
```
qurbonzoda, line 32: WDYT about the following changes:
* Change param values to "4" and "10".
* Make `list` a mutable list.
* Fill the `list` in `Setup` function
* Clear the `list` in `Teardown` function

I think the sample will look a bit more like real use case after these changes.

qurbonzoda (changes requested):

docs/writing-benchmarks.md
wldeh: I have added these in the latest commit. I imagine there is a lot of ways we can make the example more robust and similar to common real use cases.

docs/writing-benchmarks.md
```diff
@@ -0,0 +1,77 @@
...
+    }
+}
+```
+
+### Explaining the Annotations
+
+#### @State
+
+The `@State` annotation, when set to `Scope.Benchmark`, is applied to a class to represent that this class is responsible for holding the state or data for your benchmark tests. This class instance is shared among all benchmark threads, creating a consistent environment across all warmups, iterations, and measured executions. It helps in managing stateful operations in a multi-threaded context, providing an efficient way of handling state that's thread-safe and consistent across multiple runs. This ensures accurate and reliable benchmarking results, as the shared state remains the same throughout all the tests, preventing any discrepancies that could affect the outcomes. JMH provides other scope options besides `Scope.Benchmark` including `Scope.Thread`, which means each thread gets its own state instance, and `Scope.Group`, where the state is shared within a benchmark thread group.  In our snippet, the ExampleBenchmark class uses the @State(Scope.Benchmark) annotation, indicating that the state in this class is shared across all benchmark threads.
```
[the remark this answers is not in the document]
wldeh: Got it!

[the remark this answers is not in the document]
wldeh: Do you believe it is necessary to mention Java as well?

README.md
```diff
@@ -315,24 +315,77 @@ Note: Kotlin/WASM is an experimental compilation target for Kotlin. It may be dr
...
+    fun benchmarkMethod() {
+
+    }
+
+    @TearDown
+    fun cleanup() {
+
+    }
+}
+```
```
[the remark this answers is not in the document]
wldeh: Agreed. I revamped this in the latest commit.

docs/writing-benchmarks.md
wldeh: I prefer this as well.

wldeh: Their is Level.Trial and Level.Iteration however I am assuming this is only for Kotlin/JVM and Java? 

I am getting this error when I try to use them in Kotlin/JS 
`Unresolved reference: Level`
`Too many arguments for public constructor Setup() defined in kotlinx.benchmark.Setup`

wldeh: In any distinct way compared to the @teardown annotation, not that I am aware of. Could you provide more context?

docs/writing-benchmarks.md
qurbonzoda: If you create a new list in `Setup`, it does not make any sense to clear it in `Teardown`.

qurbonzoda: > I imagine there is a lot of ways we can make the example more robust and similar to common real use cases.

Any options to make the example more realistic are welcome.

docs/writing-benchmarks.md
```diff
@@ -0,0 +1,78 @@
...
...se connections, or preparing any other resources your benchmark requires. In the Kotlin/JVM and Java targets, the `@Setup` annotation can operate on two levels - `Iteration` and `Trial`. In `Iteration` level, the setup method is run before each benchmark iteration, ensuring a consistent state across all iterations. On the other hand, `Trial` level execution runs the setup method once for the entire set of benchmark method iterations, suitable when state modifications are part of the benchmark itself. Specify the level using `@Setup(Level.Trial)` or `@Setup(Level.Iteration)`; if not defined, it defaults to `Level.Trial`. The key point to remember is that the `@Setup` method's execution time is not included in the final benchmark results - the timer starts only when the `@Benchmark` method begins. This makes `@Setup` an ideal place for initialization tasks that should not impact the timing results of your benchmark. By using the `@Setup` annotation, you ensure consistency across all executions of your benchmark, providing accurate and reliable results. In the provided example, the `@Setup` annotation is used to prepare a list containing integers from 0 to a specified parameter value.
```
qurbonzoda, line 45: On what level does the `Setup` operate in other targets?

docs/writing-benchmarks.md
```diff
@@ -0,0 +1,78 @@
...
+
+#### @TearDown
+
+The `@TearDown` annotation is used to denote a method that's executed after the benchmarking method(s), typically responsible for cleaning up or deallocating any resources or conditions that were initialized or consumed during the benchmark. For instance, if your benchmark creates temporary files or opens network connections, the method marked with `@TearDown` is where you would put the code to delete those files or close those connections. The `@TearDown` annotation helps you avoid performance bias and ensure the proper maintenance of resources and the preparation of a clean environment for the next run. In our example, the `cleanup` function annotated with `@TearDown` is used to clear the list of integers after each benchmark iteration, ensuring that each test starts with the same initial state.
```
qurbonzoda, line 54: As you have already described in `Setup`, creating temporary files and opening network connections should be conducted there, not in a benchmark method.

docs/writing-benchmarks.md
```diff
@@ -0,0 +1,78 @@
...
+
+#### @BenchmarkMode
+
+The `@BenchmarkMode` annotation sets the mode of operation for the benchmark. Applying the `@BenchmarkMode` annotation requires specifying a mode from the `Mode` enum, which includes several options. `Mode.Throughput` measures the raw throughput of your code in terms of the number of operations it can perform per unit of time, such as operations per second. When this mode is selected, the toolkit runs your code in a loop for a set duration, then calculates and reports the operation throughput. Conversely, `Mode.AverageTime` is used when you're more interested in the average time it takes to execute an operation. Here, the toolkit runs your code multiple times, records the time each operation takes, and calculates and reports the average time. Without an explicit `@BenchmarkMode` annotation, the toolkit defaults to `Mode.Throughput`. In our example, `@BenchmarkMode(Mode.Throughput)` is used, meaning the benchmark focuses on the number of times the benchmark method can be executed per unit of time.
```
qurbonzoda, line 58: Could you please clarify how you inferred the information about how the toolkit runs in different modes?

qurbonzoda (changes requested):

docs/writing-benchmarks.md
qurbonzoda: According to the JMH documentation `@Setup` can be applied only in a `@State` class:
https://github.com/openjdk/jmh/blob/09c78d5d0752ffc409e64ca4cabe9dc7b96704d7/jmh-core/src/main/java/org/openjdk/jmh/annotations/Setup.java#L35

qurbonzoda: Yes, only for Kotlin/JVM.
For other targets `Setup` and `Teardown` work on Trial level. And there is no way to adjust it.

docs/writing-benchmarks.md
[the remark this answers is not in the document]
qurbonzoda: No, I don't think so.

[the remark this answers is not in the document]
wldeh: Understood

docs/writing-benchmarks.md
wldeh: Yes, the benchmark methods must always be public as you noted.
https://github.com/openjdk/jmh/blob/09c78d5d0752ffc409e64ca4cabe9dc7b96704d7/jmh-core/src/main/java/org/openjdk/jmh/annotations/Benchmark.java#L53

docs/writing-benchmarks.md
wldeh: Noted ✅

wldeh: @qurbonzoda Thanks for clarifying. Curious as to where this ranks on our priority list in terms of adding support for adjustable levels in other targets? Other than that, the documentation has been updated to reflect the current state.

docs/writing-benchmarks.md
wldeh: Got it. I updated the @Teardown method to print to the console for now and will brainstorm ways to potentially revamp the example to reflect a more coherent use case that utilizes the libraries annotations effectively.

docs/writing-benchmarks.md
```diff
@@ -0,0 +1,78 @@
+## Writing Benchmarks
+
+To get started, let's look at a simple example:
+
+```kotlin
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@Warmup(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 1, time = 1, timeUnit = TimeUnit.SECONDS)
+@Fork(1)
```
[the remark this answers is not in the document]
wldeh: Noted. I did not attempt to run it in Kotlin/JS. I decided to move forward by removing the @Fork annotation from the example and modifying the annotation description accordingly.

docs/writing-benchmarks.md
```diff
@@ -0,0 +1,78 @@
+## Writing Benchmarks
+
+To get started, let's look at a simple example:
+
+```kotlin
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@Warmup(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 1, time = 1, timeUnit = TimeUnit.SECONDS)
```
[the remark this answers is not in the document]
wldeh: Thank you for pointing this out. I have updated them accordingly.

docs/writing-benchmarks.md
```diff
@@ -0,0 +1,78 @@
...
+
+#### @Param
+
+The `@Param` annotation is used to pass different parameters to your benchmark method. It allows you to run the same benchmark method with different input values, so you can see how these variations affect performance. The values you provide for the `@Param` annotation are the different inputs you want to use in your benchmark test. The benchmark will run once for each provided value. In our example, `@Param` annotation is used with values '1' and '2', which means the benchmarkMethod will be executed twice, once with the `param` value as '1' and then with '2'. This could serve to help in understanding how the size of the input list impacts the time it takes to sum all the integers in the list.
```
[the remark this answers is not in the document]
wldeh: Thank you for pointing this out!

docs/writing-benchmarks.md
wldeh: Setup defaults to @Level.Trial on all targets as you noted earlier. I will update the documentation to reflect this info. https://github.com/openjdk/jmh/blob/09c78d5d0752ffc409e64ca4cabe9dc7b96704d7/jmh-core/src/main/java/org/openjdk/jmh/annotations/Setup.java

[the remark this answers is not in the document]
wldeh: Thank you for pointing this out. I have added it in the latest commit.

docs/writing-benchmarks.md
wldeh: Noted!

docs/writing-benchmarks.md
wldeh: In attempt to provide a quality annotation description, I left oversimplifications (not based on code) about how the toolkits operates. I have removed them in the latest commit.

docs/writing-benchmarks.md
```diff
@@ -0,0 +1,78 @@
...
...ass to represent that this class is responsible for holding the state or data for your benchmark tests. This class instance is shared among all benchmark threads, creating a consistent environment across all warmups, iterations, and measured executions. It helps in managing stateful operations in a multi-threaded context, providing an efficient way of handling state that's thread-safe and consistent across multiple runs. This ensures accurate and reliable benchmarking results, as the shared state remains the same throughout all the tests, preventing any discrepancies that could affect the outcomes. In the Kotlin/JVM and Java taregts, apart from `Scope.Benchmark`, you also have access to `Scope.Thread` and `Scope.Group`. `Scope.Thread` ensures each thread has its own unique state instance, while `Scope.Group` allows state sharing within a benchmark thread group. However, for other target environments, only `Scope.Benchmark` is supported, limiting the scope options to a single instance that is shared among all threads. In our snippet, the ExampleBenchmark class uses the @State(Scope.Benchmark) annotation, indicating that the state in this class is shared across all benchmark threads.
```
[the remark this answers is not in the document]
wldeh: Agreed. I have added this information and more in the latest commit.

docs/writing-benchmarks.md
```diff
@@ -0,0 +1,81 @@
...
...ch as Just-In-Time compilation, class loading, and garbage collection. These can significantly impact the performance of our code. However, these influences might differ from one run to another, leading to inconsistent or misleading benchmark results if multiple benchmarks are executed within the same JVM process. By triggering the JVM to fork for each benchmark, these JVM-specific factors are eliminated from affecting the benchmark results, providing a clean and independent environment for each benchmark, enhancing the reliability and comparability of results. The value you assign to the `@Fork` annotation determines the number of separate JVM processes initiated for each benchmark. If this annotation is not specified, JVMs are not forked and all benchmarks are run in the same JVM process. Repeating the benchmark across multiple JVMs and averaging the results gives a more accurate representation of typical performance and accommodates for variability possibly caused by different JVM startups. The `@Fork(1)` annotation for exmaple, indicates that each benchmark test should run in one separate JVM process, thus ensuring an isolated and reliable testing environment for each test run.
```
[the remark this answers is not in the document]
wldeh: Yes. I have fixed this inaccuracy in both PR's based on the JMH defaults https://github.com/openjdk/jmh/blob/master/jmh-core/src/main/java/org/openjdk/jmh/runner/Defaults.java#L88 . I will look to prevent oversights like this in the future.

docs/writing-benchmarks.md
qurbonzoda: I don't think we have enough time to implement this.

qurbonzoda: Definitely after documentation and integration tests

qurbonzoda:
Merged manually. Thanks!

PR #132, opened 2023-07-21
Avoid using object pinning in native blackhole

### LINKED ISSUE #114: Benchmarking results are unstable for native target due to object pinning done by the Blackhole
Benchmarking results on native platforms are unstable and measured performance degrades from iteration to iteration.

### Issue 
While running some benchmarks for the native target I spotted the problem: performance degrades from iteration to iteration.
The root cause is how the Blackhole is implemented for native: its `consume` methods pins a value: https://github.com/Kotlin/kotlinx-benchmark/blob/master/runtime/nativeMain/src/kotlinx/benchmark/NativeBlackhole.kt#L7
```
actual class Blackhole {
    actual fun consume(obj: Any?) {
        obj?.pin()
    }
...
```

References pinned by the blackhole are never unpinned and that exposes additional work for native GC as every pinned reference is a GC root.

The issue could be reproduced with the KMP example from the repository, but I added an additional configuration (https://github.com/Kotlin/kotlinx-benchmark/commit/204c22353122dcbacf098980ec19ef1ce73db7ab) to make the issue more obvious:

```
$ ./gradlew nativePinningIssueBenchmark
… test.NativeTestBenchmark.sqrtBenchmark
Warm-up #0: 9,217.20 ops/ms
Warm-up #1: 8,086.55 ops/ms
Warm-up #2: 6,463.68 ops/ms
Warm-up #3: 2,530.98 ops/ms
Warm-up #4: 2,163.17 ops/ms
Warm-up #5: 1,716.82 ops/ms
Warm-up #6: 1,456.48 ops/ms
Warm-up #7: 9,395.89 ops/ms
Warm-up #8: 7,259.49 ops/ms
Warm-up #9: 7,420.73 ops/ms
Iteration #0: 450.845 ops/ms
Iteration #1: 395.228 ops/ms
Iteration #2: 319.273 ops/ms
...
```
As you can see, measured time drifts significantly, and it degrades with each iteration.

Here's a flame graph built from dtrace profiling results:

### Suggested fix

To avoid the issue NativeBlackhole may use the same approach as the JMH's blackhole and pin objects conditionally with a condition that is always false, like:
```
   init {
         val rnd = Random(getTimeMillis())
         param1 = rnd.nextInt()
         param2 = param1 + 1
     }

     actual fun consume(obj: Any?) {
         if (param1 xor param2 == 0) {
             obj?.pin()
         }
     }
```
Here's a commit with a suggested workaround: https://github.com/fzhinkin/kotlinx-benchmark/commit/439e2acfd7a6d096c1eeb1240385c9b386f3732e

Note that actual pinning is not necessary and we can save the `consume`'s argument to some public field instead.

With that change benchmarks execution results are stable:
```
$ ./gradlew nativePinningIssueBenchmark
… test.NativeTestBenchmark.sqrtBenchmark
Warm-up #0: 35,376.2 ops/ms
Warm-up #1: 39,050.2 ops/ms
Warm-up #2: 39,947.6 ops/ms
Warm-up #3: 39,730.9 ops/ms
Warm-up #4: 39,609.2 ops/ms
Warm-up #5: 40,083.4 ops/ms
Warm-up #6: 40,185.6 ops/ms
Warm-up #7: 39,982.4 ops/ms
Warm-up #8: 39,912.6 ops/ms
Warm-up #9: 40,048.8 ops/ms
Iteration #0: 60,159.8 ops/ms
Iteration #1: 60,255.4 ops/ms
Iteration #2: 60,261.2 ops/ms
Iteration #3: 60,538.9 ops/ms
Iteration #4: 60,8401.0 ops/ms
```

fzhinkin:
Reimplemented native backhole `consume` methods using:
- for primitive types, inline asm call accepting the consumed value and preventing it from being DCEd;
- for objects, an effectively unreachable code that creates a weak reference for an object and publishes it.

There's no way to obtain the address of an object without interacting with the runtime, that's why I chose not to use an inline asm call to prevent the DCE for objects.

For primitive types, the implementation follows the approach used by Google Benchmark's `DoNotOptimize` [implementation](https://github.com/google/benchmark/blame/27d64a2351b98d48dd5b18c75ff536982a4ce26a/include/benchmark/benchmark.h#L464) and Meta's Folly `compiler_must_not_elide` [implementation](https://github.com/facebook/folly/blob/bea73e236ce14e36bca7cc28e644ff928856adb3/folly/lang/Hint-inl.h#L94).

**Please note** that Google warn users when they use `DoNotOptimize` with rvalue or const reference as an optimizer may apply some optimization for such an argument (https://github.com/google/benchmark/commit/53df805dc820d720734d6538032f6b73f31f13ba). The motivating example was about hoisting an expression out of a loop, see more details in the comments

Verified stability and correctness using the reproducer mentioned here: https://github.com/Kotlin/kotlinx-benchmark/issues/114#issuecomment-1645353693

With old implementation of the blackhole results looked like this:
```
… org.example.SampleBenchmark.explicitBlackholeBoolean
Warm-up #0: 479.799 ops/ms
Warm-up #1: 199.445 ops/ms
Warm-up #2: 153.165 ops/ms
Iteration #0: 129.017 ops/ms
Iteration #1: 109.637 ops/ms
Iteration #2: 95.7437 ops/ms
Iteration #3: 84.6977 ops/ms
Iteration #4: 76.3860 ops/ms
Iteration #5: 69.5794 ops/ms
Iteration #6: 63.7696 ops/ms
Iteration #7: 58.6253 ops/ms
Iteration #8: 54.5784 ops/ms
Iteration #9: 51.2170 ops/ms
  Success:   ~ 79.3251 ops/ms ±19%
```

With the new one:
```
… org.example.SampleBenchmark.explicitBlackholeBoolean
Warm-up #0: 36,434.4 ops/ms
Warm-up #1: 35,841.9 ops/ms
Warm-up #2: 35,866.9 ops/ms
Iteration #0: 95,282.2 ops/ms
Iteration #1: 93,515.7 ops/ms
Iteration #2: 93,098.8 ops/ms
Iteration #3: 90,363.6 ops/ms
Iteration #4: 93,872.9 ops/ms
Iteration #5: 95,240.3 ops/ms
Iteration #6: 94,687.6 ops/ms
Iteration #7: 93,960.4 ops/ms
Iteration #8: 94,340.9 ops/ms
Iteration #9: 94,610.5 ops/ms
  Success:   ~ 93,897.3 ops/ms ±0.94%
```

Fixes #114


fzhinkin:
Regarding the issue w/ applicability of some optimizations given the current implementation w/ an inline asm only reading the data, here's a simplified example based on the description of a problem in the gbenchmark's PR:
https://godbolt.org/z/7P5dr61zP. Unless the `key` could be modified by the inline asm, an optimizer can hoist dependent computations (`val` in this case) out of the loop.

I've checked if it could be an issue for K/N, it didn't inline the consume call and was able to hoist loop invariants out of the inner loop:
```
    @Benchmark
    fun blackholeTest(bh: Blackhole) {
        for (i in 0 until 1000) {
            val key = nextLcg()
            for (j in 0 until 500) {
                bh.consume(key)
                val v = (key shr 17) + 0xdead
                bh.consume(v)
            }
        }
    }
```
Corresponding assembly:
```
000000010006adb4 <_kfun:kotlinx.benchmark.generated.test.NativeTestBenchmark_Descriptor.$blackholeTest$FUNCTION_REFERENCE$4.invoke#internal>:
<prologue>
10006addc: c8 fe df 08  ldarb   w8, [x22]
10006ade0: 48 00 00 36  tbz     w8, #0, 0x10006ade8 <_kfun:kotlinx.benchmark.generated.test.NativeTestBenchmark_Descriptor.$blackholeTest$FUNCTION_REFERENCE$4.invoke#internal+0x34>
10006ade4: 94 33 00 94  bl      0x100077c34 <__ZN6kotlin2mm26SuspendIfRequestedSlowPathEv>
10006ade8: 17 00 80 52  mov     w23, #0
10006adec: b8 c1 8c 52  mov     w24, #26125
10006adf0: 38 03 a0 72  movk    w24, #25, lsl #16
10006adf4: f9 6b 9e 52  mov     w25, #62303
10006adf8: d9 8d a7 72  movk    w25, #15470, lsl #16
10006adfc: ba d5 9b 52  mov     w26, #57005
10006ae00: 04 00 00 14  b       0x10006ae10 <_kfun:kotlinx.benchmark.generated.test.NativeTestBenchmark_Descriptor.$blackholeTest$FUNCTION_REFERENCE$4.invoke#internal+0x5c>
;;
;; outer loop:
;;
10006ae04: f7 06 00 11  add     w23, w23, #1
10006ae08: ff a2 0f 71  cmp     w23, #1000
10006ae0c: 80 02 00 54  b.eq    0x10006ae5c <_kfun:kotlinx.benchmark.generated.test.NativeTestBenchmark_Descriptor.$blackholeTest$FUNCTION_REFERENCE$4.invoke#internal+0xa8>
10006ae10: c8 fe df 08  ldarb   w8, [x22]
10006ae14: 48 00 00 36  tbz     w8, #0, 0x10006ae1c <_kfun:kotlinx.benchmark.generated.test.NativeTestBenchmark_Descriptor.$blackholeTest$FUNCTION_REFERENCE$4.invoke#internal+0x68>
10006ae18: 87 33 00 94  bl      0x100077c34 <__ZN6kotlin2mm26SuspendIfRequestedSlowPathEv>
10006ae1c: 68 12 40 b9  ldr     w8, [x19, #16]
;;
;; val key = nextLcg()
;;
10006ae20: 14 65 18 1b  madd    w20, w8, w24, w25
10006ae24: 74 12 00 b9  str     w20, [x19, #16]
;;
;; val v = (key shr 17) + 0xdead
;;
10006ae28: 55 47 94 0b  add     w21, w26, w20, asr #17
10006ae2c: 9b 3e 80 52  mov     w27, #500
10006ae30: 07 00 00 14  b       0x10006ae4c <_kfun:kotlinx.benchmark.generated.test.NativeTestBenchmark_Descriptor.$blackholeTest$FUNCTION_REFERENCE$4.invoke#internal+0x98>
;;
;; inner loop
;;
10006ae34: e0 03 14 aa  mov     x0, x20
10006ae38: 66 db ff 97  bl      0x100061bd0 <_kfun:kotlinx.benchmark.Blackhole#consume(kotlin.Int){}>
10006ae3c: e0 03 15 aa  mov     x0, x21
10006ae40: 64 db ff 97  bl      0x100061bd0 <_kfun:kotlinx.benchmark.Blackhole#consume(kotlin.Int){}>
10006ae44: 7b 07 00 71  subs    w27, w27, #1
10006ae48: e0 fd ff 54  b.eq    0x10006ae04 <_kfun:kotlinx.benchmark.generated.test.NativeTestBenchmark_Descriptor.$blackholeTest$FUNCTION_REFERENCE$4.invoke#internal+0x50>
10006ae4c: c8 fe df 08  ldarb   w8, [x22]
10006ae50: 28 ff 07 36  tbz     w8, #0, 0x10006ae34 <_kfun:kotlinx.benchmark.generated.test.NativeTestBenchmark_Descriptor.$blackholeTest$FUNCTION_REFERENCE$4.invoke#internal+0x80>
10006ae54: 78 33 00 94  bl      0x100077c34 <__ZN6kotlin2mm26SuspendIfRequestedSlowPathEv>
10006ae58: f7 ff ff 17  b       0x10006ae34 <_kfun:kotlinx.benchmark.generated.test.NativeTestBenchmark_Descriptor.$blackholeTest$FUNCTION_REFERENCE$4.invoke#internal+0x80>
<epilogue>
10006ae74: c0 03 5f d6  ret
```

So it seems like there's no way to prevent hoisting unless we can potentially modify a consumed value. But it'll require different API.

runtime/nativeMain/src/kotlinx/benchmark/NativeBlackhole.kt
```diff
@@ -1,35 +1,61 @@
...
     actual fun consume(obj: Any?) {
-        obj?.pin()
+        // https://en.wikipedia.org/wiki/Linear_congruential_generator
+        val next = lcg * 1664525 + 1013904223
+        if (next == 0) {
+            if (obj !== null) {
+                ref = WeakReference(obj)
+            }
+        }
+        lcg = next
```
qwwdfsad, line 31: I'm a bit concerned it is a full memory (seq consistent) barriered write on each `consume` that potentially can affect measurement precision.

Could you please elaborate a bit on why the conditional write is not enough, taking into account the fact that `lcg` already is `volatile`?

runtime/nativeMain/src/kotlinx/benchmark/NativeBlackhole.kt
```diff
@@ -1,35 +1,61 @@
...
+import kotlin.concurrent.Volatile
+import kotlin.native.ref.WeakReference
+import kotlin.random.Random
+import kotlinx.benchmark.impl.*
 
+@OptIn(ExperimentalStdlibApi::class)
 actual class Blackhole {
+    /*
+     * For primitive types, DCE may be omitted by creating an artificial use via inline asm.
+     * For objects, there is no cheap way to obtain something that could be passed to a native function with such
```
qwwdfsad, line 14: Worth asking `@Pavel Kunyavskiy` about this one.
AFAIR, there was _some_ way (kinda internal-ish) to cast `Any` to `long*`, and then we can do the same trick as with primitives (first ensuring that it does prevent e.g. scalarization)

runtime/nativeMain/src/kotlinx/benchmark/NativeBlackhole.kt
fzhinkin: The store should be eliminated, it's just a mistake. I'll rewrite this code.

runtime/nativeMain/src/kotlinx/benchmark/NativeBlackhole.kt
fzhinkin: Well, there's not dedicated API to do that, but there's `identityHashCode` (and I totally forgot about its existence) which is effectively `reinterpret_cast<uintptr_t>(this)`.

runtime/nativeMain/src/kotlinx/benchmark/NativeBlackhole.kt
```diff
@@ -1,35 +1,89 @@
...
+
+    actual inline fun consume(s: Short) {
+       consume(s.toInt())
     }
-    actual fun consume(i: Int) {
-        i.pin()
+
+    actual inline fun consume(i: Int) {
+        // To ensure that i's value will not be removed by optimizations like dead code elimination,
+        // it's value is compares with two value i0 and i1, such that i1 = i0 + 1.
```
fzhinkin, line 52:
```suggestion
        // its value is compares with two value i0 and i1, such that i1 = i0 + 1.
```

runtime/nativeMain/src/kotlinx/benchmark/NativeBlackhole.kt
```diff
@@ -1,35 +1,89 @@
...
+        // As long as i0 and i1 are different, the following condition should not ever be met
+        // and as the branch following it (note that if it is executed, then NPE will happen).
+        // To ensure that at least i0 value will be loaded on every call, it was annotated with Volatile.
+        //
+        // This approach has one drawback: in general, it should be compiled to a code with two branch instructions,
+        // and performance characteristics of a benchmark may not be stable if consumed value is sometimes equal to i0.
+        // In practice, there is almost no effect on the measured performance and
+        // the difference is withing error margin.
```
fzhinkin, line 60:
```suggestion
        // the difference is within error margin.
```

runtime/nativeMain/src/kotlinx/benchmark/NativeBlackhole.kt
```diff
@@ -1,35 +1,89 @@
...
+
+    @Volatile
+    var d0 = Random.nextDouble()
+    var d1 = d0 + 1.0
+
+    @Volatile
+    var bh: Blackhole? = null
+
+    actual inline fun consume(obj: Any?) {
+        val hc = obj?.identityHashCode() ?: 0
```
qwwdfsad, line 30: Worth dropping a comment this one is not a virtual call and is always a memory address

qwwdfsad, line 30: Also, we can split null and non-null paths into two methods if it makes any sense

qwwdfsad (approved):
Nice job!

runtime/nativeMain/src/kotlinx/benchmark/NativeBlackhole.kt
fzhinkin: Done

fzhinkin: Well, `identityHashCode` is defined for `Any?` and `Returns: 0 for null object, identity hash code otherwise.`, so I removed all the extra null checks.
Not sure if a separate path really needed, especially if it differs from `consume(Int)` (it may lead to different performance depending on how frequently we receive nulls).

qwwdfsad: Even better, thanks

whyoleg:
@fzhinkin could you update benchmark results in PR description with latest changes?
Also overall full description could be updated and summarised based on comments, as implementation now is different as to what is proposed initially (just for future readers :) )

PR #135, opened 2023-07-27
Remove the kotlin.mpp.enableGranularSourceSetsMetadata from gradle.pr…

erokhins:
…operties

This property is set to true since Kotlin ~1.6, so it is already enabled. But in 1.9.20 usage of this property is an error


qwwdfsad:
Ticket: https://youtrack.jetbrains.com/issue/KT-59304/Elevate-warning-to-Error-for-pre-HMPP-flags

qwwdfsad (approved):

mshishkina (approved):

PR #139, opened 2023-08-11
Fix the failing OptionsValidationTest

qurbonzoda:
The expected message was incorrect. Also, rectify the list of valid options.


qurbonzoda:
Merged into master manually.

PR #150, opened 2023-09-02
Advance Gradle version to 8.0.2

igoriakovlev:


examples/kotlin-multiplatform/build.gradle
```diff
@@ -13,12 +13,20 @@ allOpen {
...
-    js('jsIr', IR) { nodejs() }
-    js('jsIrBuiltIn', IR) { nodejs() }
+    js('jsIr', IR) {
+        nodejs()
+        attributes.attribute(distinguishAttribute, 'jsIr')
+    }
+    js('jsIrBuiltIn', IR) {
+        nodejs()
+        attributes.attribute(distinguishAttribute, 'jsIrBuiltIn')
+    }
```
qurbonzoda, line 29: I was thinking about fixing the "targets are not distinguishable" issue by having a single JS target and adding a compilation. e.g.
```
    js(IR) {
        nodejs()
        compilations.create('builtInExecutor') {
            associateWith(compilations.main)
        }
    }
```

Considering the fact that declaring multiple targets of the same kind [won't be supported soon](https://youtrack.jetbrains.com/issue/KT-59316), this seems like a better long-term solution.

WDYT?

qurbonzoda (changes requested):
Could you please clarify why the Gradle version needs to be updated?

igoriakovlev:
> Could you please clarify why the Gradle version needs to be updated?

Honestly I do not remember why I need it, but I decided it could be useful. We can close this MR if you think it is not necessary this time.

PR #154, opened 2023-10-09
Changes for 2.0 migration

mglukhikh:
This changes are needed to be able to build kotlinx.benchmarks in Aggregate with 2.0. According to decisions around related YT issues, the changes should be Ok (some error suppressions removed + Gradle dependencies made more clear).

### DIFF
--- a/runtime/jvmMain/src/kotlinx/benchmark/JvmBenchmarkAnnotations.kt
+++ b/runtime/jvmMain/src/kotlinx/benchmark/JvmBenchmarkAnnotations.kt
@@ -20,7 +20,6 @@ actual typealias OutputTimeUnit = org.openjdk.jmh.annotations.OutputTimeUnit
 @Suppress("ACTUAL_WITHOUT_EXPECT")
 actual typealias BenchmarkTimeUnit = java.util.concurrent.TimeUnit
 
-@Suppress("ACTUAL_ANNOTATION_CONFLICTING_DEFAULT_ARGUMENT_VALUE", "NO_ACTUAL_CLASS_MEMBER_FOR_EXPECTED_CLASS")
 actual typealias Warmup = org.openjdk.jmh.annotations.Warmup
 
 @Suppress("ACTUAL_ANNOTATION_CONFLICTING_DEFAULT_ARGUMENT_VALUE")
--- a/examples/kotlin-kts/build.gradle.kts
+++ b/examples/kotlin-kts/build.gradle.kts
@@ -28,6 +28,7 @@ dependencies {
 tasks.withType<KotlinCompile> {
     kotlinOptions {
         jvmTarget = "1.8"
+        languageVersion = rootProject.properties["kotlin_language_version"].toString()
     }
 }
 
--- a/examples/kotlin-multiplatform/build.gradle
+++ b/examples/kotlin-multiplatform/build.gradle
@@ -28,6 +28,7 @@ kotlin {
     sourceSets.all {
         languageSettings {
             progressiveMode = true
+            languageVersion = kotlin_language_version
         }
     }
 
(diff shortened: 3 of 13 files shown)


woainikk:
@qurbonzoda can we take these changes to develop branch?

runtime/jvmMain/src/kotlinx/benchmark/JvmBenchmarkAnnotations.kt
```diff
@@ -10,8 +10,8 @@ actual typealias TearDown = org.openjdk.jmh.annotations.TearDown
 
 actual typealias Benchmark = org.openjdk.jmh.annotations.Benchmark
 
-@Suppress("NO_ACTUAL_CLASS_MEMBER_FOR_EXPECTED_CLASS")
-actual typealias BenchmarkMode = org.openjdk.jmh.annotations.BenchmarkMode
+@Target(AnnotationTarget.CLASS)
+actual annotation class BenchmarkMode(actual vararg val value: Mode)
```
qurbonzoda, line 14: This workaround is temporary and will be fixed before going into master, right?
I found related open issue that will hopefully be fixed in 2.0: https://youtrack.jetbrains.com/issue/KT-61100

This workaround changes how benchmarks run in JVM. Basically, the annotation is ignored in JVM now.

runtime/jvmMain/src/kotlinx/benchmark/JvmBenchmarkAnnotations.kt
```diff
@@ -20,11 +20,9 @@ actual typealias OutputTimeUnit = org.openjdk.jmh.annotations.OutputTimeUnit
...
 actual typealias BenchmarkTimeUnit = java.util.concurrent.TimeUnit
 
-@Suppress("ACTUAL_ANNOTATION_CONFLICTING_DEFAULT_ARGUMENT_VALUE", "NO_ACTUAL_CLASS_MEMBER_FOR_EXPECTED_CLASS")
 actual typealias Warmup = org.openjdk.jmh.annotations.Warmup
 
 @Suppress("ACTUAL_ANNOTATION_CONFLICTING_DEFAULT_ARGUMENT_VALUE")
 actual typealias Measurement = org.openjdk.jmh.annotations.Measurement
 
-@Suppress("NO_ACTUAL_CLASS_MEMBER_FOR_EXPECTED_CLASS")
-actual typealias Param = org.openjdk.jmh.annotations.Param
```
qurbonzoda, line 30: The same here: https://github.com/Kotlin/kotlinx-benchmark/pull/154/files#r1350778685

runtime/jvmMain/src/kotlinx/benchmark/JvmBenchmarkAnnotations.kt
mglukhikh: > This workaround changes how benchmarks run in JVM. Basically, the annotation is ignored in JVM now.

In these circumstances I'd suggest that we shouldn't merge this PR right now. I asked MPP team again about some design around this problem.

mglukhikh:
@qurbonzoda What do you think about two other commits (https://github.com/Kotlin/kotlinx-benchmark/pull/154/commits/9c5fe9dc43868fbb46bff2c343639fcff95f9dbd, https://github.com/Kotlin/kotlinx-benchmark/pull/154/commits/e71060c0dff8661d70486f7638de916f32569290)? Do they also damage something or they can be accepted?

qurbonzoda:
The two other commits do not damage anything. The arguments added to the `@Warmup` annotation will have effect only in JVM. Other platforms will ignore them. However, the changes won't break or alter behavior of existing code.
It makes sense to document that the arguments are ignored in non-JVM platforms. In the future we will consider supporting them in other platforms as well.

mglukhikh:
For @qurbonzoda and others: after some consideration I decided to drop the questionable commit from here, with the intention to merge two other commits. About https://youtrack.jetbrains.com/issue/KT-61100/K2-wrong-type-for-value-parameter-of-java-annotation-constructor, I will try to fix it this or next week (or ask somebody other about it).

qurbonzoda (approved):

woainikk:
can we take these changes to develop branch directly?

ilya-g (approved):

woainikk (approved):

PR #158, opened 2023-10-18
A set of Gradle script changes to provide K2 compatibility

mglukhikh:


examples/kotlin-multiplatform/build.gradle
```diff
@@ -17,8 +17,11 @@ kotlin {
     jvm {
         compilations.create('benchmark') { associateWith(compilations.main) }
     }
-    js('jsIr', IR) { nodejs() }
-    js('jsIrBuiltIn', IR) { nodejs() }
+    js(IR) {
+        nodejs()
+        compilations.create("defaultExecutor") { associateWith(compilations.main) }
```
qurbonzoda, line 22: Minor: This compilation can be eliminated. Instead, the `js` target can be registered for benchmarking with the default executor.

qurbonzoda (approved):

PR #163, opened 2023-11-28
Advance Kotlin version to 1.9.21

qurbonzoda:


plugin/main/src/kotlinx/benchmark/gradle/BenchmarksPlugin.kt
```diff
@@ -37,8 +37,8 @@ class BenchmarksPlugin : Plugin<Project> {
 
         plugins.withType(org.jetbrains.kotlin.gradle.plugin.KotlinBasePlugin::class.java) { kotlinPlugin ->
             logger.info("Detected Kotlin plugin version '${kotlinPlugin.pluginVersion}'")
-            if (!getKotlinVersion(kotlinPlugin.pluginVersion).isAtLeast(1, 9, 0)) {
-                logger.error("JetBrains Gradle Benchmarks plugin requires Kotlin version 1.9.0 or higher")
+            if (!getKotlinVersion(kotlinPlugin.pluginVersion).isAtLeast(1, 9, 20)) {
+                logger.error("JetBrains Gradle Benchmarks plugin requires Kotlin version 1.9.20 or higher")
```
fzhinkin, line 41: README states that the minimum required version is 1.7.20. Should it be updated as well?

plugin/main/src/kotlinx/benchmark/gradle/JsMultiplatformTasks.kt
```diff
@@ -46,7 +46,7 @@ private fun Project.createJsBenchmarkCompileTask(target: JsBenchmarkTarget): Kot
             )
         }
 
-        compileKotlinTask.apply {
+        compileTaskProvider.get().apply {
```
ilya-g, line 49: `configure { }` instead of `get().apply { }`?

plugin/main/src/kotlinx/benchmark/gradle/BenchmarksPlugin.kt
fzhinkin: Discussed that the docs will be updated later.

fzhinkin (approved):

plugin/main/src/kotlinx/benchmark/gradle/BenchmarksPlugin.kt
qurbonzoda: Here is the PR to update README: https://github.com/Kotlin/kotlinx-benchmark/pull/167

plugin/main/src/kotlinx/benchmark/gradle/JsMultiplatformTasks.kt
qurbonzoda: Done

igoriakovlev (approved):

ilya-g (approved):

PR #164, opened 2023-11-28
Enable default template hierarchy

qurbonzoda:
Followup to https://github.com/Kotlin/kotlinx-benchmark/pull/163

### DIFF
--- a/examples/kotlin-multiplatform/build.gradle
+++ b/examples/kotlin-multiplatform/build.gradle
@@ -1,5 +1,3 @@
-import org.jetbrains.kotlin.konan.target.HostManager
-import org.jetbrains.kotlin.konan.target.KonanTarget
 import kotlinx.benchmark.gradle.JsBenchmarksExecutor
 
 plugins {
@@ -23,10 +21,14 @@ kotlin {
         compilations.create("builtInExecutor") { associateWith(compilations.main) }
     }
     wasm('wasmJs') { d8() }
-    if (HostManager.host == KonanTarget.MACOS_X64.INSTANCE) macosX64('native')
-    if (HostManager.host == KonanTarget.MACOS_ARM64.INSTANCE) macosArm64('native')
-    if (HostManager.hostIsLinux) linuxX64('native')
-    if (HostManager.hostIsMingw) mingwX64('native')
+
+    // Native targets
+    macosX64()
+    macosArm64()
+    linuxX64()
+    mingwX64()
+
+    applyDefaultHierarchyTemplate()
 
     sourceSets.configureEach {
         languageSettings {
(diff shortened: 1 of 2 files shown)


examples/kotlin-multiplatform/build.gradle
```diff
@@ -121,6 +121,6 @@ benchmark {
             jsBenchmarksExecutor = JsBenchmarksExecutor.BuiltIn
         }
         register("wasmJs")
-        register("native")
+        register("konan")
```
fzhinkin, line 124: Why can't we leave it `native`?

fzhinkin: Thanks for the explanation!

examples/kotlin-multiplatform/build.gradle
```diff
@@ -23,10 +23,12 @@ kotlin {
...
     }
     wasm('wasmJs') { d8() }
-    if (HostManager.host == KonanTarget.MACOS_X64.INSTANCE) macosX64('native')
-    if (HostManager.host == KonanTarget.MACOS_ARM64.INSTANCE) macosArm64('native')
-    if (HostManager.hostIsLinux) linuxX64('native')
-    if (HostManager.hostIsMingw) mingwX64('native')
+    if (HostManager.host == KonanTarget.MACOS_X64.INSTANCE) macosX64('konan')
+    if (HostManager.host == KonanTarget.MACOS_ARM64.INSTANCE) macosArm64('konan')
+    if (HostManager.hostIsLinux) linuxX64('konan')
+    if (HostManager.hostIsMingw) mingwX64('konan')
```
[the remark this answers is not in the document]
ilya-g: Since this is an example, it should be similar to how user multiplatform projects are set up, and they definitely do not use this trick with HostManager anymore.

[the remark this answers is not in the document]
qurbonzoda: Done. 
By the way, it turns out the plugin does not create benchmark tasks for non-host targets. It just reports that benchmarks for those targets cannot be run in current host, e.g.:
`Skipping benchmarks for 'linuxX64' because they cannot be run on current OS: Expected macos_x64, but was linux_x64`
So the situation is better.

fzhinkin (approved):

examples/kotlin-multiplatform/build.gradle
```diff
@@ -23,10 +21,14 @@ kotlin {
...
-    if (HostManager.hostIsLinux) linuxX64('native')
-    if (HostManager.hostIsMingw) mingwX64('native')
+
+    // Native targets
+    macosX64()
+    macosArm64()
+    linuxX64()
+    mingwX64()
+
+    applyDefaultHierarchyTemplate()
```
ilya-g, line 31: I believe we should strive that the project contains no custom `dependsOn` so that the template is applied automatically.

ilya-g (approved):

PR #165, opened 2023-11-28
Enable allWarningsAsErrors compiler option for :runtime and :examples:kotlin-multiplatform projects

qurbonzoda:


fzhinkin (approved):

runtime/build.gradle
```diff
@@ -108,3 +107,11 @@ if (project.findProperty("publication_repository") == "space") {
 tasks.withType(KotlinNativeCompile).matching { it.name.toLowerCase().contains("test") }.configureEach {
     it.dependsOn(tasks.withType(Sign))
 }
+
+tasks.withType(KotlinNativeCompile).configureEach {
+    compilerOptions.freeCompilerArgs.addAll(
+            "-opt-in=kotlin.experimental.ExperimentalNativeApi",
+            "-opt-in=kotlin.native.runtime.NativeRuntimeApi",
+            "-opt-in=kotlinx.cinterop.ExperimentalForeignApi",
```
ilya-g, line 115: Maybe you could continue using `languageSettings.optIn` DSL but move it to `nativeMain` source set?

ilya-g (approved):

runtime/build.gradle
qurbonzoda: Unfortunately, this approach doesn't work. The opt-ins are not passed to the dependent source sets automatically.
This causes:
```
Inconsistent settings for Kotlin source sets: 'androidNativeArm32Main' depends on 'nativeMain'
  'androidNativeArm32Main': set of opt-in annotations in use is []
  'nativeMain': set of opt-in annotations in use is [kotlin.experimental.ExperimentalNativeApi, kotlin.native.runtime.NativeRuntimeApi, kotlinx.cinterop.ExperimentalForeignApi]
  The dependent source set must use all opt-in annotations that its dependency uses.
```

examples/kotlin-multiplatform/build.gradle
```diff
@@ -30,6 +30,12 @@ kotlin {
 
     applyDefaultHierarchyTemplate()
 
+    targets.configureEach {
+        compilations.configureEach {
+            kotlinOptions.allWarningsAsErrors = true
```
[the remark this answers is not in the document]
qurbonzoda: Done.
But configuring `compilerOptions` in Groovy is not as handy as in Kotlin.
It is not possible to call `compilerOptions.configure {}` function because of resolution ambiguity between `fun configure(configuration: T.() -> Unit)` and `fun configure(configuration: Action<T>)`.
So the `compilerOptions.options` needs to be configured directly.

PR #172, opened 2023-12-12
Remove use of forUseAtConfigurationTime

liutikas:
This method has been deprecated and calls to System.getProperty are handle correctly

See https://docs.gradle.org/8.5/userguide/upgrading_version_7.html#for_use_at_configuration_time_deprecation

### DIFF
--- a/plugin/main/src/kotlinx/benchmark/gradle/Utils.kt
+++ b/plugin/main/src/kotlinx/benchmark/gradle/Utils.kt
@@ -261,13 +261,8 @@ internal val Gradle.isConfigurationCacheAvailable
     } ?: false
 
 internal fun Project.getSystemProperty(key: String): String? {
-    return if (gradle.isConfigurationCacheAvailable) {
-        providers.systemProperty(key).forUseAtConfigurationTime().orNull
-    } else {
-        System.getProperty(key)
-    }
+    return System.getProperty(key)
 }
-
 fun Project.javaCompilerProvider(): Provider<JavaCompiler> = provider {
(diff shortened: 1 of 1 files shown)


plugin/main/src/kotlinx/benchmark/gradle/Utils.kt
```diff
@@ -256,9 +256,5 @@ internal val Gradle.isConfigurationCacheAvailable
     } ?: false
 
 internal fun Project.getSystemProperty(key: String): String? {
-    return if (gradle.isConfigurationCacheAvailable) {
-        providers.systemProperty(key).forUseAtConfigurationTime().orNull
```
qurbonzoda, line 260: According to the migration guide:
> The Provider based APIs are still the recommended way to connect external values to task inputs for maximum configuration cache reuse.

Should `providers.systemProperty(key).orNull` be used instead of `System.getProperty(key)`?

plugin/main/src/kotlinx/benchmark/gradle/Utils.kt
```diff
@@ -256,9 +256,5 @@ internal val Gradle.isConfigurationCacheAvailable
     } ?: false
 
 internal fun Project.getSystemProperty(key: String): String? {
-    return if (gradle.isConfigurationCacheAvailable) {
```
qurbonzoda, line 259: Is the `Gradle.isConfigurationCacheAvailable` property used anywhere else? If not, the property declaration above can be removed as well?

qurbonzoda (changes requested):

PR #170, opened 2023-11-29
Add a guide to tasks created by the plugin

qurbonzoda:
Followup to https://github.com/Kotlin/kotlinx-benchmark/pull/168

The changes were developed within GSoC 2023 program by our mentee @wldeh.

PR #175, opened 2023-12-26
Add a guide to creating a separate module for benchmarks

qurbonzoda:
Followup to https://github.com/Kotlin/kotlinx-benchmark/pull/170

The changes were developed within GSoC 2023 program by our mentee @wldeh.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ kotlinx-benchmark is a toolkit for running benchmarks for multiplatform code wri
   - [Writing Benchmarks](#writing-benchmarks)
   - [Running Benchmarks](#running-benchmarks)
   - [Benchmark Configuration Profiles](#benchmark-configuration-profiles)
-  - [Separate source sets for benchmarks](#separate-source-sets-for-benchmarks)
+  - [Separate source set for benchmarks](#separate-source-set-for-benchmarks)
 - [Examples](#examples)
 - [Contributing](#contributing)
 
@@ -423,10 +423,10 @@ benchmark {
 
 Refer to our [comprehensive guide](docs/configuration-options.md) to learn about configuration options and how they affect benchmark execution.
 
-### Separate source sets for benchmarks
+### Separate source set for benchmarks
 
 Often you want to have benchmarks in the same project, but separated from main code, much like tests.
-Refer to our [detailed documentation](docs/separate-benchmark-source-set.md) on configuring your project to add a separate source set for benchmarks.
+Refer to our [detailed documentation](docs/separate-benchmark-source-set.md) on configuring your project to set up a separate source set for benchmarks.
 
 ## Examples
 
--- a/docs/separate-benchmark-source-set.md
+++ b/docs/separate-benchmark-source-set.md
@@ -1,0 +1,112 @@
+# Setting Up a Separate Source Set for Benchmarks
+
+This guide will walk you through the process of establishing a dedicated source set for benchmarks within your Kotlin project. 
+This approach is especially beneficial when you are integrating benchmarks into an existing project. 
+Here are a couple of advantages of doing so:
+
+1. **Flexibility**: Setting up a separate source set allows you to manage your benchmarking code independently. You can compile, test, and run benchmarks without impacting your main source code.
+
+2. **Organization**: It helps maintain a clean and organized project structure. Segregating benchmarks from the main code makes it easier to navigate and locate specific code segments.
+
+## Step-by-step Setup Guide
+
+Below are the step-by-step instructions to set up a separate source set for benchmarks in both Kotlin Multiplatform and Kotlin/JVM projects:
+
+### Kotlin Multiplatform Project
+
+Follow these steps to set up a separate source set for benchmarks:
+
+1. **Define New Compilation**
+
+    Start by creating a new compilation in your target of choice (e.g. jvm, js, native, wasm etc.). 
+    In this example, we're associating the new compilation `benchmark` with the `main` compilation of the `jvm` target. 
+    This association allows the benchmark compilation to access the internal API of the main compilation, 
+    which is particularly useful when benchmarks need to measure the performance of specific components 
+    or functionalities within the main codebase.
+
+    ```kotlin
+    // build.gradle.kts
+    kotlin { 
+        jvm { 
+            compilations.create('benchmark') {
+                associateWith(compilations.main)
+            }
+        }
+    }
+    ```
+
+2. **Register Benchmark Compilation**
+
+    Register your new benchmark compilation using its default source set name. 
+    In this instance, `jvmBenchmark` is the name of the default source set of the `benchmark` compilation.
+
+    ```kotlin
+    // build.gradle.kts
+    benchmark {
+        targets {
+            register("jvmBenchmark")
+        }
+    }
+    ```
+
+3. **Add Benchmarks**
+
+    Place your benchmark code into the default source set of the benchmark compilation. 
+    The default source set can also depend on other source sets containing benchmarks. 
+    This way you can share benchmarks between multiple benchmark compilations. 
+    Refer to our [writing benchmarks guide](docs/writing-benchmarks.md) for a comprehensive guide on writing benchmarks.
+
+For additional information, refer to the [Kotlin documentation on creating a custom compilation](https://kotlinlang.org/docs/multiplatform-configure-compilations.html#create-a-custom-compilation).
+and the [documentation on associating compiler tasks](https://kotlinlang.org/docs/gradle-configure-project.html#associate-compiler-tasks).
+[Here is a sample Kotlin Multiplatform project](/examples/kotlin-multiplatform) with a separate compilation for benchmarks.
+
+### Kotlin/JVM Project
+
+Set up a separate benchmark source set by following these simple steps:
+
+1. **Define Source Set**
+
+    Begin by defining a new source set. We'll use `benchmark` as the name for the source set.
+
+    ```kotlin
+    // build.gradle.kts
+    sourceSets {
+        create("benchmark")
+    }
+    ```
+
+2. **Propagate Dependencies**
+
+    Next, propagate dependencies and output from the `main` source set to your `benchmark` source set. 
+    This ensures the `benchmark` source set has access to classes and resources from the `main` source set.
+
+    ```kotlin
+    // build.gradle.kts
+    dependencies {
+        add("benchmarkImplementation", sourceSets.main.get().output + sourceSets.main.get().runtimeClasspath)
+    }
+    ```
+
+    You can also add output and `compileClasspath` from `sourceSets.test` in the same way 
+    if you wish to reuse some of the test infrastructure.
+
+3. **Register Benchmark Source Set**
+
+    Register your benchmark source set. This informs the kotlinx-benchmark tool 
+    that benchmarks reside within this source set and need to be executed accordingly.
+
+    ```kotlin
+    // build.gradle.kts
+    benchmark {
+        targets { 
+            register("benchmark")
+        }
+    }
+    ```
+
+4. **Add Benchmarks**
+
+   Place your benchmark code into the benchmark source set.
+   Refer to our [writing benchmarks guide](docs/writing-benchmarks.md) for a comprehensive guide on writing benchmarks.
+
+[Here is a sample Kotlin/JVM project](/examples/kotlin) with custom source set for benchmarks.


docs/separate-benchmark-module.md
```diff
@@ -0,0 +1,102 @@
...
+    benchmark {
+        targets {
+            register("jvmBenchmark")
+        }
+    }
+    ```
+
+For additional information, refer to the [Kotlin documentation on creating a custom compilation](https://kotlinlang.org/docs/multiplatform-configure-compilations.html#create-a-custom-compilation).
+and the [documentation on associating compiler tasks](https://kotlinlang.org/docs/gradle-configure-project.html#associate-compiler-tasks).
+[Here is a sample Kotlin Multiplatform project](/examples/kotlin-multiplatform) with a separate compilation for benchmarks.
```
fzhinkin, line 56: It confuses a bit that examples in this documents use Kotlin-script only, but referred sample projects use Groovy build scripts.
Maybe it's worth replacing Groovy scripts in sample projects with Kotlin scripts afterward.

docs/separate-benchmark-module.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Step-by-step Setup Guide
+
+Below are the step-by-step instructions to set up a separate module for benchmarks in both Kotlin Multiplatform and Kotlin/JVM projects:
+
+### Kotlin Multiplatform Project
+
+Follow these steps to set up a separate compilation for benchmarks:
+
+1. **Define New Compilation**:
```
fzhinkin, line 21: Here, "Define New Compilation" is like a section heading, so I'm not sure if it should ends with a colon.

docs/separate-benchmark-module.md
```diff
@@ -0,0 +1,102 @@
...
+    }
+    ```
+
+For additional information, refer to the [Kotlin documentation on creating a custom compilation](https://kotlinlang.org/docs/multiplatform-configure-compilations.html#create-a-custom-compilation).
+and the [documentation on associating compiler tasks](https://kotlinlang.org/docs/gradle-configure-project.html#associate-compiler-tasks).
+[Here is a sample Kotlin Multiplatform project](/examples/kotlin-multiplatform) with a separate compilation for benchmarks.
+
+### Kotlin/JVM Project
+
+Set up a separate benchmark source set by following these simple steps:
```
fzhinkin, line 60: According the title, the document is about creating a separate module for benchmarks, but this sections describes how to add a benchmarks source set instead.

For me as a reader it raises a few questions:
1) Can I create a separate module w/ benchmarks for K/JVM project? Or it's only for KMP and in JVM-only project I can only add benchmarks into a separate source set?
2) If I can add a separate module into K/JVM project too, do I always need a main source set and separate benchmarks source set?

docs/separate-benchmark-module.md
```diff
@@ -0,0 +1,102 @@
...
+
+Below are the step-by-step instructions to set up a separate module for benchmarks in both Kotlin Multiplatform and Kotlin/JVM projects:
+
+### Kotlin Multiplatform Project
+
+Follow these steps to set up a separate compilation for benchmarks:
+
+1. **Define New Compilation**:
+
+    Start by creating a new compilation in your target of choice (e.g. jvm, js, native, wasm etc.).
```
fzhinkin, line 23: Is it something a user always have to do when creating a separate module with benchmarks or it's some advanced feature? If it's the latter then when should a user configure it?
I've added several separate benchmarking modules (that are using `kotlinx-benchmarks`) into several separate projects, but I have never created separate compilations. It was always "add a module, apply plugin, add dependency on a module under test, register targets".

docs/separate-benchmark-module.md
```diff
@@ -0,0 +1,102 @@
+# Creating Separate Module for Benchmarks
+
+This guide will walk you through the process of creating a separate module for benchmarks in your Kotlin project. 
+Creating a separate module for benchmarks is especially beneficial when you are integrating benchmarks into an existing project. 
+Here are a few advantages of doing so:
+
+1. **Organization**: It helps maintain a clean and organized project structure. Segregating benchmarks from the main code makes it easier to navigate and locate specific code segments.
```
fzhinkin, line 7: But to some degree that could be also achieved by using a separate source set, isn't it?

docs/separate-benchmark-module.md
```diff
@@ -0,0 +1,102 @@
...
+
+2. **Isolation**: Separating benchmarks ensures that the benchmarking code does not interfere with your main code or test code. This isolation guarantees accurate measurements without unintentional side effects.
```
fzhinkin, line 9: But that could be also achieved using a separate source set, not a module.
It's unclear why there should be issues with the measurements accuracy when benchmarks are placed along with main sources.

docs/separate-benchmark-module.md
```diff
@@ -0,0 +1,102 @@
...
+
+3. **Flexibility**: Creating a separate module allows you to manage your benchmarking code independently. You can compile, test, and run benchmarks without impacting your main source code.
```
fzhinkin, line 11: That's a fair point and I think the list should starts with it.

fzhinkin (changes requested):

docs/separate-benchmark-module.md
qurbonzoda: We already have a PR to replace Groovy with Kotlin in build scripts of the sample projects.
The PR still needs some work, but the final goal is to update the sample projects.

docs/separate-benchmark-module.md
qurbonzoda: `kotlinx-benchmark` supports both Kotlin Multiplatform and Kotlin/JVM projects.

In Kotlin Multiplatform (K/M) project:
* If the benchmark code is integrated into the main compilation of Kotlin targets (for instance, when the project is specifically dedicated to benchmarks), Kotlin targets should be registered as benchmark targets. This is documented in the updated README.
* If one prefers to separate benchmark code from the main compilation, separate compilations should be created and registered as benchmark targets, using the names of their main source set. The steps for this are outlined in the guide.

In Kotlin/JVM project:
* If the benchmark code is part of the main source set (for example, in a project dedicated to benchmarks), the main source set should be registered as the benchmark target. A separate guide will be provided for this scenario.
* If one wishes to segregate benchmark code from the main source set, a separate source set should be created and registered as the benchmark target, as described in this guide.

It's important to note that in K/M projects, compilations are used, while in K/JVM projects, source sets are employed, as there is no separate Kotlin compilation in K/JVM projects. The term "module" is used in this guide to encompass both "compilation" and "source set" contexts. I would appreciate any suggestions to improve the wording.

qurbonzoda: The described setup might not be the most logical, but that is how the library works. See [BenchmarksExtension.kt](https://github.com/Kotlin/kotlinx-benchmark/blob/3f32f2d9d9732a2a01d231134c52e6554428b519/plugin/main/src/kotlinx/benchmark/gradle/BenchmarksExtension.kt#L42) for more details on how `kotlinx-benchmark` handles the registration of benchmark targets.

fzhinkin: I guess the main source of the confusion is

> The term "module" is used in this guide to encompass both "compilation" and "source set" contexts. I would appreciate any suggestions to improve the wording.

Idea uses the term "module" as a synonym for Gradle sub-project, so when I hear "module" the only thing I'm thinking of is sub-projects. 

The term "module" has multiple meaning and the document does not specify that here the word "module" means a compilation or a source set, depending on project type. 
As the word "module" is used only a couple of times and referred in the main README, perhaps it's worth to abstain from using it at all and instead write something like "separating benchmark sources from the main project sources"?

Then, the preamble could be rewritten to something like:
```
This guide will walk you through the process of separating benchmarks from the main sources in your Kotlin project. 
That is especially beneficial when you are integrating benchmarks into an existing project.
 ...
 <the list of benefits>
 ...
Depending on a project type (Multiplatform or JVM), different approaches should be employed to keep benchmark sources aside of the main project sources. The guide below will walk you through required steps for each of the supported project types.
```

I'm not going to claim that it's the best wording, but it seems to be less ambiguous both in terms of the intention and the approach required to achieve it.

docs/separate-benchmark-module.md
fzhinkin: As we figured it out, the term "module" is about compilations, so one has to set it up. ;)

docs/separate-benchmark-module.md
fzhinkin: And it is the source set.

docs/separate-benchmark-module.md
qurbonzoda: Yes, it is considered an advanced feature. It is needed when one wants to use/access/benchmark internal API of an existing project.
As I described [above](https://github.com/Kotlin/kotlinx-benchmark/pull/175#discussion_r1436951614), the basic scenario is to create a separate project (or subproject) that is dedicated to benchmarks.

docs/separate-benchmark-module.md
qurbonzoda: It can be. However, in K/Multiplatform projects each source set has to be included into a compilation, thus there is no way to avoid compilations.
Including the source set with benchmarks into the main compilation means:
* The benchmark code is also published together with the main code.
* Dependencies of the benchmark code will be in the list of dependencies of the published artifact.

Perhaps it's possible to avoid these problems, but it needs more advanced configurations of the project.

The recommended (by the library) way is to set up the benchmark compilation similar to the test compilation, which is separated from the main compilation.

docs/separate-benchmark-module.md
qurbonzoda: > But that could be also achieved using a separate source set, not a module.

https://github.com/Kotlin/kotlinx-benchmark/pull/175#discussion_r1437011229

> t's unclear why there should be issues with the measurements accuracy when benchmarks are placed along with main sources.

Agree. I will remove this point.

docs/separate-benchmark-module.md
qurbonzoda: Done

docs/separate-benchmark-module.md
qurbonzoda: I don't have strong opinion in this regard. But it does look better without colon.
Colons used in section headings are removed now.

docs/separate-benchmark-module.md
qurbonzoda: I've changed the guide to use "source set" throughout. The reason is eventually the benchmark code in placed into source sets in both Kotlin Multiplatform and Kotlin/JVM projects. Please take a look if the new wording is any good?

README.md
```diff
@@ -425,10 +425,10 @@ benchmark {
 
 Refer to our [comprehensive guide](docs/configuration-options.md) to learn about configuration options and how they affect benchmark execution.
 
-### Separate module for benchmarks
+### Separate source set for benchmarks
 
 Often you want to have benchmarks in the same project, but separated from main code, much like tests.
-Refer to our [detailed documentation](docs/separate-benchmark-module.md) on configuring your project to add a separate module for benchmarks.
+Refer to our [detailed documentation](docs/separate-benchmark-source-set.md) on configuring your project to set up a separate source for benchmarks.
```
fzhinkin, line 431:
```suggestion
Refer to our [detailed documentation](docs/separate-benchmark-source-set.md) on configuring your project to set up a separate source set for benchmarks.
```

🤓

fzhinkin (approved):

docs/separate-benchmark-module.md
fzhinkin: Thanks for addressing the issues. Everything looks good to me!

docs/separate-benchmark-source-set.md
```diff
@@ -0,0 +1,112 @@
...
+        targets {
+            register("jvmBenchmark")
+        }
+    }
+    ```
+
+3. **Add Benchmarks**
+
+    Place your benchmark code into the default source set of the benchmark compilation. 
+    The default source set can also depend on other source sets with benchmarks.
```
Goooler, line 55:
```suggestion
    The default source set can also depend on other source sets containing benchmarks. 
```

docs/separate-benchmark-source-set.md
```diff
@@ -0,0 +1,112 @@
...
+    }
+    ```
+
+    You can also add output and `compileClasspath` from `sourceSets.test` in the same way 
+    if you wish to reuse some of the test infrastructure.
+
+3. **Register Benchmark Source Set**
+
+    Register your benchmark source set. This informs the kotlinx-benchmark tool 
+    that benchmarks reside within this source set and need to be executed accordingly.
```
Goooler, line 96:
```suggestion
    about benchmarks reside within this source set and need to be executed accordingly.
```

Goooler (commented):
And the title should be `Add a guide for creating` instead of `to`?

docs/separate-benchmark-source-set.md
qurbonzoda: I am not sure if the change is grammatically correct.

qurbonzoda:
> And the title should be `Add a guide for creating` instead of `to`?

I surfed internet and it seems both prepositions are fine to use in this case.

PR #184, opened 2024-01-26
Add a guide to writing benchmarking

qurbonzoda:
Followup to https://github.com/Kotlin/kotlinx-benchmark/pull/182

The changes were developed within GSoC 2023 program by our mentee @wldeh.

### DIFF
--- a/docs/writing-benchmarks.md
+++ b/docs/writing-benchmarks.md
@@ -1,0 +1,239 @@
+# Writing Benchmarks
+
+If you're familiar with the Java Microbenchmark Harness (JMH) toolkit, you'll find that the `kotlinx-benchmark`
+library shares a similar approach to crafting benchmarks. This compatibility allows you to seamlessly run your
+JMH benchmarks written in Kotlin on various platforms with minimal, if any, modifications.
+
+Like JMH, kotlinx-benchmark is annotation-based, meaning you configure benchmark execution behavior using annotations.
+The library then extracts metadata provided through annotations to generate code that benchmarks the specified code
+in the desired manner.
+
+To get started, let's examine a simple example of a multiplatform benchmark:
+
+```kotlin
+import kotlinx.benchmark.*
+
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(BenchmarkTimeUnit.MILLISECONDS)
+@Warmup(iterations = 10, time = 500, timeUnit = BenchmarkTimeUnit.MILLISECONDS)
+@Measurement(iterations = 20, time = 1, timeUnit = BenchmarkTimeUnit.SECONDS)
+@State(Scope.Benchmark)
+class ExampleBenchmark {
+
+    // Parameterizes the benchmark to run with different list sizes
+    @Param("4", "10")
+    var size: Int = 0
+
+    private val list = ArrayList<Int>()
+
+    // Prepares the test environment before each benchmark run
+    @Setup
+    fun prepare() {
+        for (i in 0..<size) {
+            list.add(i)
+        }
+    }
+
+    // Cleans up resources after each benchmark run
+    @TearDown
+    fun cleanup() {
+        list.clear()
+    }
+
+    // The actual benchmark method
+    @Benchmark
+    fun benchmarkMethod(): Int {
+        return list.sum()
+    }
+}
+```
+
+**Example Description**:
+This example tests the speed of summing numbers in an `ArrayList`. We evaluate this operation with lists
+of 4 and 10 numbers to understand the method's performance with different list sizes.
+
+## Explaining the Annotations
+
+The following annotations are available to define and fine-tune your benchmarks.
+
+### @State
+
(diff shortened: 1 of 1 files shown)


docs/writing-benchmarks.md
```diff
@@ -0,0 +1,236 @@
...
+    for (e in myList) {
+        bh.consume(e)
+    }
+}
+```
+
+By consuming results, you signal to the compiler that these computations are significant and shouldn't be optimized away.
+
+For a deeper dive into `Blackhole` and its nuances in JVM, you can refer to:
+- [Official Javadocs](https://javadoc.io/static/org.openjdk.jmh/jmh-core/1.23/org/openjdk/jmh/infra/Blackhole.html)
```
fzhinkin, line 235:
```suggestion
- [Official Javadocs](https://javadoc.io/doc/org.openjdk.jmh/jmh-core/latest/org/openjdk/jmh/infra/Blackhole.html)
```

docs/writing-benchmarks.md
```diff
@@ -0,0 +1,236 @@
...
+This example tests the speed of summing numbers in an `ArrayList`. We evaluate this operation with lists
+of 4 and 10 numbers to understand the method's performance with different list sizes.
+
+## Explaining the Annotations
+
+The following annotations are available to define and fine-tune your benchmarks.
+
+### @State
+
+The `@State` annotation is used to mark benchmark classes.
```
fzhinkin, line 58: This paragraph gives a good overview of the `State`'s purpose but also leaves an impression that that purpose changes depending on a target platform.

I think we can write that it always `specifies the extent to which the state object is shared among the worker threads` and by default, it's mandatory. However, right now, the multi-threaded execution is only supported on JVM and there the scope has some other values. Also, when writing a JVM-only benchmark, that annotation could be omitted.

docs/writing-benchmarks.md
```diff
@@ -0,0 +1,236 @@
...
+
+### @BenchmarkMode
+
+The `@BenchmarkMode` annotation sets the mode of operation for the benchmark.
+
+Applying the `@BenchmarkMode` annotation requires specifying a mode from the `Mode` enum.
+In Kotlin/JVM, the `Mode` enum has several options, including `SingleShotTime`.
+
+Refer to [JMH documentation of Mode](https://javadoc.io/doc/org.openjdk.jmh/jmh-core/latest/org/openjdk/jmh/annotations/Mode.html)
+for details about available options. In other targets, only `Throughput` and `AverageTime` are available.
```
fzhinkin, line 132: As with `State`, I think it's worth describing the general case first (covering `Throughput` and `AverageTime` modes) and then mentioning that on JVM the set of available modes is broader.

docs/writing-benchmarks.md
```diff
@@ -0,0 +1,236 @@
...
+in the desired manner.
+
+To get started, let's examine a simple example of a multiplatform benchmark:
+
+```kotlin
+import kotlinx.benchmark.*
+
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@Warmup(iterations = 20, time = 1, timeUnit = TimeUnit.SECONDS)
```
fzhinkin, line 18: I think it's worth using some values different from what is passed to `@Masurement` to show that these two annotations don't have to be in sync.

fzhinkin (changes requested):

docs/writing-benchmarks.md
```diff
@@ -0,0 +1,236 @@
...
+`Mode.Throughput` measures the raw throughput of your code in terms of the number of operations it can perform per unit
+of time, such as operations per second. `Mode.AverageTime` is used when you're more interested in the average time it
+takes to execute an operation. Without an explicit `@BenchmarkMode` annotation, the toolkit defaults to `Mode.Throughput`.
+
+The annotation is put at the enclosing class and has the effect over all `@Benchmark` methods in the class.
+In Kotlin/JVM, it may be put at `@Benchmark` method to have effect on that method only.
+Refer to [JMH documentation of @BenchmarkMode](https://javadoc.io/doc/org.openjdk.jmh/jmh-core/latest/org/openjdk/jmh/annotations/BenchmarkMode.html)
+for details about the effect of the annotation in Kotlin/JVM.
+
+In our example, `@BenchmarkMode(Mode.Throughput)` is used, meaning the benchmark focuses on the number of times the
```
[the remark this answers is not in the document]
qurbonzoda: It does. Now `AverageTime` is used. PTAL.

docs/writing-benchmarks.md
qurbonzoda: Done. PTAL if it's better now.

docs/writing-benchmarks.md
qurbonzoda: Done.

docs/writing-benchmarks.md
qurbonzoda: Agree. Done.

fzhinkin (approved):
LGTM

Issue #130, opened 2023-07-19
Support benchmarks having @State annotations in parent classes


fzhinkin:
`kotlinx-benchmark` requires benchmark classes to be annotated with `@State`.

It seems reasonable to expect that when there is a base class annotated with `@State` providing common routines and state for subclasses, and then it should be enough to simply extend that class w/o explicitly annotating leaf class w/ `@State`.

Here's an example of such code reuse via inheritance from `kotlinx-io` [benchmarks](https://github.com/Kotlin/kotlinx-io/blob/master/benchmarks/src/commonMain/kotlin/BufferOps.kt#L12):
```kotlin
@State(Scope.Benchmark)
abstract class BufferRWBenchmarkBase {
    protected val buffer = Buffer()
    @Setup fun setupBuffers() { /* some setup common for all subclasses */ }
}

open class ByteBenchmark : BufferRWBenchmarkBase() {
    @Benchmark fun benchmark(): Byte {
        buffer.writeByte(0x42)
        return buffer.readByte()
    }
}

open class ShortBenchmark : BufferRWBenchmarkBase() {
    @Benchmark fun benchmark(): Short {
        buffer.writeShort(42)
        return buffer.readShort()
    }
}

...
```

Also, it would be nice to verify that if benchmark-class and its superclass are both annotated w/ `@State` with the same scope. The condition would be always true now, but may no longer hold in the future.

My expectations regarding State inheritance are based on how it works in JMH, and because that's a different framework it not necessarily should be true for `kotlinx-benchmark`. However, if the proposal will be rejected, it would be nice to print warning for the classes having `@Benchmark`-methods, but not annotated w/ the `@State`.

07jasjeet:
Hi! can I pick this one up?

qurbonzoda:
@07jasjeet Yes please

PR #193, opened 2024-03-13
#130: Support benchmarks having @State annotations in parent classes

07jasjeet:
## Changelog
1. Whenever a parent of a child class is directly or indirectly annotated with `@State` annotation, our child class will also be considered when generating benchmarks.
2. Supplimentary annotations like `@Measurement` `@Warmup` `@OutputTime` `@BenchmarkMode` are considered in this order: child -> top most annotated parent

**Note**: This is a phase-1 implementation, with just the basic idea laid out.

### DIFF
--- a/plugin/main/src/kotlinx/benchmark/gradle/SuiteSourceGenerator.kt
+++ b/plugin/main/src/kotlinx/benchmark/gradle/SuiteSourceGenerator.kt
@@ -103,17 +103,17 @@ class SuiteSourceGenerator(val title: String, val module: ModuleDescriptor, val 
             val classDescriptors = DescriptorUtils.getAllDescriptors(packageFragment.getMemberScope())
                 .filterIsInstance<ClassDescriptor>()
 
-            val abstractClassDescriptors = classDescriptors.filter { it.modality == Modality.ABSTRACT }
+            val inheritableClassDescriptors = classDescriptors.filter { it.modality != Modality.FINAL }
 
             // Abstract classes which are annotated directly.
-            val directlyAnnotatedAbstractClassDescriptors = abstractClassDescriptors
+            val directlyAnnotatedInheritableClassDescriptors = inheritableClassDescriptors
                 .filter { it.annotations.any { it.fqName.toString() == stateAnnotationFQN } }
 
             // Contains both directly and indirectly annotated ClassDescriptors.
-            val annotatedAbstractClassDescriptors = abstractClassDescriptors
-                .filter { abstractClass ->
-                    directlyAnnotatedAbstractClassDescriptors.forEach { annotatedAbstractClass ->
-                        if (abstractClass.isSubclassOf(annotatedAbstractClass)) {
+            val annotatedInheritableClassDescriptors = inheritableClassDescriptors
+                .filter { inheritableClass ->
+                    directlyAnnotatedInheritableClassDescriptors.forEach { annotatedInheritableClass ->
+                        if (inheritableClass.isSubclassOf(annotatedInheritableClass)) {
                             // If any benchmark class is not annotated but extended with an abstract class
                             // annotated with @State, it should be included when generating benchmarks.
                             return@filter true
@@ -122,24 +122,17 @@ class SuiteSourceGenerator(val title: String, val module: ModuleDescriptor, val 
                     return@filter false
                 }
 
-            fun ClassDescriptor.getParentAnnotated(): ClassDescriptor? {
-                annotatedAbstractClassDescriptors.forEach { annotatedAbstractClass ->
-                    if (this.isSubclassOf(annotatedAbstractClass)) {
-                        return annotatedAbstractClass
-                    }
-                }
-                return null
-            }
-
             val annotatedClassDescriptors = mutableListOf<AnnotatedClassDescriptor>()
                 .apply {
                     classDescriptors
                         .filter { it.modality != Modality.ABSTRACT }
                         .forEach {
-                            if (it.annotations.any { it.fqName.toString() == stateAnnotationFQN }) {
+                            if (it.annotations.any {
+                                annotationDescriptor -> annotationDescriptor.fqName.toString() == stateAnnotationFQN
+                            }) {
                                 add(AnnotatedClassDescriptor(it))
                             } else {
-                                val parent = it.getParentAnnotated()
+                                val parent = it.getParentAnnotated(annotatedInheritableClassDescriptors)
                                 if (parent != null) {
                                     add(AnnotatedClassDescriptor(it, parent))
                                 }
@@ -157,7 +150,22 @@ class SuiteSourceGenerator(val title: String, val module: ModuleDescriptor, val 
         }
     }
 
-    /** @param parentAnnotatedClassDescriptor if null, [original] is directly annotated.*/
+    /** @param annotatedInheritableClassDescriptors descriptors of classes which are inheritable and directly annotated.
+     * with [stateAnnotationFQN].
+     * @return Top most parent class descriptor which was annotated with [stateAnnotationFQN] and inherited by `this` class
+     * descriptor or null if none of its parent classes are not annotated.*/
+    private fun ClassDescriptor.getParentAnnotated(
+        annotatedInheritableClassDescriptors: List<ClassDescriptor>
+    ): ClassDescriptor? {
+        annotatedInheritableClassDescriptors.forEach { annotatedAbstractClass ->
+            if (this.isSubclassOf(annotatedAbstractClass)) {
+                return annotatedAbstractClass
+            }
+        }
+        return null
+    }
+
+    /** @param parentAnnotatedClassDescriptor if null, [original] is directly annotated with [stateAnnotationFQN].*/
     private data class AnnotatedClassDescriptor(
         val original: ClassDescriptor,
         val parentAnnotatedClassDescriptor: ClassDescriptor? = null


plugin/main/src/kotlinx/benchmark/gradle/SuiteSourceGenerator.kt
```diff
@@ -99,21 +100,99 @@ class SuiteSourceGenerator(val title: String, val module: ModuleDescriptor, val
...
-    private fun generateBenchmark(original: ClassDescriptor) {
+    /** @param annotatedInheritableClassDescriptors descriptors of classes which are inheritable and directly annotated.
+     * with [stateAnnotationFQN].
+     * @return Top most parent class descriptor which was annotated with [stateAnnotationFQN] and inherited by `this` class
+     * descriptor or null if none of its parent classes are not annotated.*/
+    private fun ClassDescriptor.getParentAnnotated(
+        annotatedInheritableClassDescriptors: List<ClassDescriptor>
+    ): ClassDescriptor? {
+        annotatedInheritableClassDescriptors.forEach { annotatedAbstractClass ->
+            if (this.isSubclassOf(annotatedAbstractClass)) {
```
qurbonzoda, line 161: Could it be that there are multiple superclasses for `this` class descriptor?

examples/kotlin/benchmarks/src/TestInheritedBenchmark.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.openjdk.jmh.annotations.*
+import java.util.concurrent.TimeUnit
+import kotlin.math.cos
+import kotlin.math.sqrt
+
+@State(Scope.Benchmark)
+@Fork(1)
+@Warmup(iterations = 0)
+@Measurement(iterations = 1, time = 1, timeUnit = TimeUnit.SECONDS)
+abstract class BaseBenchmark {
```
qurbonzoda, line 12: Can the base class declare benchmark methods as well?

examples/kotlin/benchmarks/src/TestInheritedBenchmark.kt
```diff
@@ -0,0 +1,34 @@
...
+
+    protected lateinit var data: TestData
+
+    @Setup
+    fun setUp() {
+        data = TestData(50.0)
+    }
+}
+
+class TestInheritedBenchmark: BaseBenchmark() {
```
qurbonzoda, line 22: Can a `TestInheritedBenchmark` subclass declare benchmark methods as well?

qurbonzoda (changes requested):

qurbonzoda:
> Note: This is a phase-1 implementation, with just the basic idea laid out.

I would appreciate it if you could clarify what exactly you think should be included in phase 1.

plugin/main/src/kotlinx/benchmark/gradle/SuiteSourceGenerator.kt
```diff
@@ -99,21 +100,99 @@ class SuiteSourceGenerator(val title: String, val module: ModuleDescriptor, val
 
     private fun processPackage(module: ModuleDescriptor, packageView: PackageViewDescriptor) {
         for (packageFragment in packageView.fragments.filter { it.module == module }) {
-            DescriptorUtils.getAllDescriptors(packageFragment.getMemberScope())
+            val classDescriptors = DescriptorUtils.getAllDescriptors(packageFragment.getMemberScope())
                 .filterIsInstance<ClassDescriptor>()
+
+            val inheritableClassDescriptors = classDescriptors.filter { it.modality != Modality.FINAL }
+
+            // Abstract classes which are annotated directly.
```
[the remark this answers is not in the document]
07jasjeet: Oops yes old comment. Will fix!

plugin/main/src/kotlinx/benchmark/gradle/SuiteSourceGenerator.kt
```diff
@@ -99,21 +100,99 @@ class SuiteSourceGenerator(val title: String, val module: ModuleDescriptor, val
 
     private fun processPackage(module: ModuleDescriptor, packageView: PackageViewDescriptor) {
         for (packageFragment in packageView.fragments.filter { it.module == module }) {
-            DescriptorUtils.getAllDescriptors(packageFragment.getMemberScope())
+            val classDescriptors = DescriptorUtils.getAllDescriptors(packageFragment.getMemberScope())
                 .filterIsInstance<ClassDescriptor>()
+
+            val inheritableClassDescriptors = classDescriptors.filter { it.modality != Modality.FINAL }
```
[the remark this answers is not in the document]
07jasjeet: Sounds good.

plugin/main/src/kotlinx/benchmark/gradle/SuiteSourceGenerator.kt
07jasjeet: I was expecting this! So as for phase-1 I was just looking out for bare bones implementation which means this here just a base setup. Currently this function picks the **first** superclass that is directly annotated. Now there are some cases:
1) Simple straight forward: Only one annotated superclass.
2) Multiple superclasses but only one is annotated.
3) Multiple superclasses and intermediaries (super class of this but sub class of the top most annotated parent class) are annotated as well.
4) Multiple superclasses and intermediaries with annotations E.x., `@Warmup`, `@Measurement` etc without `@State`.

What should be the route to follow here?
For 4th point, my suggestion is that independent annotations should be ignored. For the 3rd point, I would suggest nearest annotated parent. 1st and 2nd are easily dealt with.

examples/kotlin/benchmarks/src/TestInheritedBenchmark.kt
07jasjeet: No. As to why? -> It doesn't make sense to run the same benchmark again if two or more classes are inheriting from same non-final class.

examples/kotlin/benchmarks/src/TestInheritedBenchmark.kt
07jasjeet: Could you expand a bit on your question?

plugin/main/src/kotlinx/benchmark/gradle/SuiteSourceGenerator.kt
```diff
@@ -99,21 +100,99 @@ class SuiteSourceGenerator(val title: String, val module: ModuleDescriptor, val
...
+                .apply {
+                    classDescriptors
+                        .filter { it.modality != Modality.ABSTRACT }
+                        .forEach {
+                            if (it.annotations.any {
+                                annotationDescriptor -> annotationDescriptor.fqName.toString() == stateAnnotationFQN
+                            }) {
+                                add(AnnotatedClassDescriptor(it))
+                            } else {
+                                val parent = it.getParentAnnotated(annotatedInheritableClassDescriptors)
```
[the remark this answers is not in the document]
07jasjeet: Huh after another thought, I think `directlyAnnotatedInheritableClassDescriptors` should be used here. Will do!

plugin/main/src/kotlinx/benchmark/gradle/SuiteSourceGenerator.kt
qurbonzoda: > What should be the route to follow here?

It depends on how JMH handles this aspect. Being consistent with that library is important, since it runs the benchmarks on the JVM. Achieving consistent behavior across other platforms would be ideal. It is worth mentioning that we do not aim to replicate all of JMH's sophistication, but when something is not supported in non-JVM targets, we would like to issue a warning (or throw an exception) about it.

qurbonzoda: I would greatly appreciate it if you could check how JMH handles this case. It would greatly help us decide which route we should take.

examples/kotlin/benchmarks/src/TestInheritedBenchmark.kt
qurbonzoda: I can think of the following use case:
```
import kotlinx.benchmark.*

@State(Scope.Benchmark)
@Measurement(iterations = 7, time = 1, timeUnit = BenchmarkTimeUnit.SECONDS)
@Warmup(iterations = 7, time = 1, timeUnit = BenchmarkTimeUnit.SECONDS)
@OutputTimeUnit(BenchmarkTimeUnit.MILLISECONDS)
abstract class IterationBenchmarkBase {
    abstract val list: List<String>

    @Param("10", "100")
    var size: Int = 0

    @Benchmark
    fun iterationBenchmark(bh: Blackhole) {
        for (element in list) {
            bh.consume(element)
        }
    }
}

class ArrayListIterationBenchmark : IterationBenchmarkBase() {
    override val list: List<String>
        get() = ArrayList<String>(size).apply { fill(size) }
}

class ArrayDequeIterationBenchmark : IterationBenchmarkBase() {
    override val list: List<String>
        get() = ArrayDeque<String>(size).apply { fill(size) }
}

private fun MutableList<String>.fill(size: Int) {
    addAll((1..size).map { it.toString() })
}
```

examples/kotlin/benchmarks/src/TestInheritedBenchmark.kt
qurbonzoda: I am referring to cases of multi-hierarchy benchmark classes, within which multiple classes define benchmark methods.

PR #195, opened 2024-03-27
remove MaxPermSize from gradle.properties in test templates

adam-enko:
The option MaxPermSize option triggers an error on my machine:

```text
Caused by: org.gradle.api.GradleException: Unable to start the daemon process.
This problem might be caused by incorrect configuration of the daemon.
For example, an unrecognized jvm option is used.
Please refer to the User Manual chapter on the daemon at https://docs.gradle.org/8.0.2/userguide/gradle_daemon.html
...
Please read the following process output to find out more:
-----------------------
Unrecognized VM option 'MaxPermSize=2048m'
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

### DIFF
--- a/integration/src/test/resources/templates/kmp-with-toolchain/gradle.properties
+++ b/integration/src/test/resources/templates/kmp-with-toolchain/gradle.properties
@@ -1,1 +1,1 @@
-org.gradle.jvmargs=-Xmx2g -XX:MaxPermSize=2048m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx2g -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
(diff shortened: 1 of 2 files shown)


adam-enko:
Replacing `MaxPermSize` with `MaxMetaspaceSize` makes sense, but setting it necessary? Do the tests use too much memory? I thought it would be simpler just to remove it.

I'm not very familiar with JVM mememory management, but as I understand the tests are run using Gradle TestKit, which should inherit the JVM args of the parent build, which has has a max limit of 4g.

ALikhachev (approved):
👍 
This looks like a kind of legacy. Since Java 8, metaspace is not limited by default, and should not be modified unless there's specific reasoning behind that, letting JVM to choose the right size itself

https://docs.oracle.com/en/java/javase/11/gctuning/other-considerations.html:
> Starting with JDK 8, the permanent generation was removed and the class metadata is allocated in native memory. The amount of native memory that can be used for class metadata is by default unlimited. Use the option -XX:MaxMetaspaceSize to put an upper limit on the amount of native memory used for class metadata.
...
Specify a higher value for the option -XX:MetaspaceSize to avoid early garbage collections induced for class metadata. The amount of class metadata allocated for an application is application-dependent and general guidelines do not exist for the selection of -XX:MetaspaceSize. The default size of -XX:MetaspaceSize is platform-dependent and ranges from 12 MB to about 20 MB.

qurbonzoda (approved):

PR #196, opened 2024-03-27
Tidy build files

adam-enko:
Some spring cleaning of some project build files.

- Remove Gradle Wrapper files from the plugin included-build 

  (It's possible these are useful, if someone wants to open the plugin project independently, but that shouldn't be necessary, so these files can be removed)

- update .gitignore

- remove NOTES.txt. It hasn't been edited in 5+ years, so I'm assuming it's not important.

None of these changes are important, I'm just doing some tidying. Feel free to reject all or none of these changes.


qurbonzoda (approved):

.gitignore
```diff
@@ -1,9 +1,7 @@
-.idea
+.idea/
 *.iml
-.gradle
-.gradletasknamecache
-build
-out
-target
-local.properties
```
ALikhachev, line 8: I would suggest keeping the `local.properties` entry here as this file may be useful for overriding some KGP properties locally, e.g. for experiments or manual testing.

ALikhachev (changes requested):

.gitignore
adam-enko: Good idea, thanks. I assumed that local.properties was only for Android development.

ALikhachev (approved):

PR #200, opened 2024-04-08
format numbers with a fixed Locale

adam-enko:
Decimal/thousands separators differ per Locale. Using the default system Locale causes [FormatTests](https://github.com/Kotlin/kotlinx-benchmark/blob/ed69605e261b45ffa5629e4a18f39a0ba1756b66/runtime/commonTest/src/kotlinx/benchmark/tests/FormatTests.kt#L7-L6) failures when the default system Locale has `,` for the decimal separator and `.` as the thousands separator (which is common in Europe).

E.g.

```
Expected :83.3333 ops/us
Actual   :83,3333 ops/us
```

image


qurbonzoda (commented):
Does the tests fail in JS and WasmJs as well?
`Double.format` seems to use the host default locale in those targets.

runtime/jvmMain/src/kotlinx/benchmark/Utils.kt
```diff
@@ -1,10 +1,11 @@
 package kotlinx.benchmark
 
-import java.io.*
+import java.io.File
+import java.util.*
 
 internal actual fun Double.format(precision: Int, useGrouping: Boolean): String {
-    return if (useGrouping) "%,.0${precision}f".format(this)
-    else "%.0${precision}f".format(this)
+    return if (useGrouping) "%,.0${precision}f".format(Locale.US, this)
```
qurbonzoda, line 7: `Locale.ROOT` for a neutral locale

ALikhachev (changes requested):
Agree with both comments. It seems like there's no direct replacement for `Locale.ROOT` in the JS implementation, so perhaps makes sense to use some fixated locale that conforms the requirements

adam-enko:
> Does the tests fail in JS and WasmJs as well? `Double.format` seems to use the host default locale in those targets.

On my machine, they both pass. It was only the JVM ones that failed.

adam-enko:
> Agree with both comments. It seems like there's no direct replacement for `Locale.ROOT` in the JS implementation, so perhaps makes sense to use some fixated locale that conforms the requirements

So you mean, instead of `d.toLocaleString(undefined, { ... } )`, instead specify a locale with the separators we want? E.g. `d.toLocaleString('en-GB', { ... } )`

ALikhachev (approved):

qurbonzoda (approved):

PR #202, opened 2024-04-09
Initial restructuring, preparing for moving Generator code

adam-enko:
This is the initial restructuring, that sets up a new empty `kotlinx-benchmark-generator` subproject that will (in the next PR) contain the generator code. (Setting this now, before moving code, allows us to evaluate the structure independently of the code.)

Part of [KT-66949](https://youtrack.jetbrains.com/issue/KT-66949/Split-kotlinx-benchmark-code-generation-into-a-separate-subproject)

#### Summary 

- Move `runtime` into `kotlinx-benchmark-core/kotlinx-benchmark-runtime`.

  The subproject was renamed, so that it's easier to reason able the project layout, publishing is easier to follow (the artifact name matches the directory name),.

  The new `kotlinx-benchmark-core` subdirectory, which will contain kotlinx-benchmark components used by the Gradle plugin at runtime, for better organisation.
- Create empty 'generator' subproject, that will (eventually) contain the generator code.


qurbonzoda (changes requested):
I suggest we merge the changes into a different branch (other than master) until we have a more clear understanding of the new structure. I am also eager to see what the `generator` project will include. Mostly `SuiteSourceGenerator.kt`?

qurbonzoda:
The concerns raised by @whyoleg resonate with me. Indeed, there is no dependency between the `plugin` and `runtime` components. Given that we will have only three components for the foreseeable future, it seems logical to place them all in the root directory. Alternatively, to distinguish them from other modules such as `docs` and `examples`, we could place them in a `core/` directory. WDYT?

adam-enko:
> But `kotlinx-benchmark-runtime` is not used by Gradle plugin runtime, rather it's used by user code runtime.

Yes, that's right. Sorry, it was unclear. I propose that the 'core' Benchmark dependencies are used at runtime, either by KGP or the Benchmark plugin, and so it makes sense to group them together.

> I suggest we merge the changes into a different branch (other than master) until we have a more clear understanding of the new structure.

I'll make a new branch 👍 

> I am also eager to see what the `generator` project will include. Mostly `SuiteSourceGenerator.kt`?

Yes, that's right. It will also contain some of the code used in the tasks to launch the generator for each Kotlin target. The end result will be that the generator project will expose 4 functions (for generating JVM, Native, JS, and Wasm).

> we could place them in a `core/` directory. WDYT?

Sure, that makes sense to me!

If it helps, I'd like to see a layout similar to [Dokka](https://github.com/Kotlin/dokka/) (which has a `dokka-subprojects` for 'core' dependencies, and a `dokka-runner` with the various plugins), or [Kotest](https://github.com/kotest/kotest) (which also has `kotest-runner` and misc `kotest-*` directories for the runtime dependencies).

qurbonzoda:
> I'd like to see a layout similar to [Dokka](https://github.com/Kotlin/dokka/) or [Kotest](https://github.com/kotest/kotest)

The layout of Dokka and Kotest is indeed very well-organized. However, for kx-benchmark in its current state, adopting a similar layout might be overkill. We currently have only one plugin, and there are no plans to add another runner in the foreseeable future. Should we add new modules in the future, we will consider adopting a similar layout.

whyoleg:
> I propose that the 'core' Benchmark dependencies are used at runtime, either by KGP or the Benchmark plugin, and so it makes sense to group them together.

for me it's not really make sense, could you elaborate on why do you think like this?
Comparing to Dokka, here the situation is different IMO. We could say, that in Dokka we have clear separation between 2 parts: Dokka Engine/Analysis (what is in `dokka-subprojects`) and Dokka Runners (Gradle/Maven plugins, CLI). Here in `benchmarks` we do have 3 parts: runtime (what is needed at runtime of benchmark to setup/execute benchmarks), generator (which theoretically, could be just a compiler/KSP plugin, generates additional sources which uses `runtime` and should be added to benchmark runtime) and plugin (Gradle Plugin, or in future multiple plugins/cli/etc, which orchestrates build so that `generator` will know what and where to generate)

And so for me, all those 3 parts are different things and there is no need to group `runtime` and `generator` at all. IMO, the only grouping that makes sense is separation of artefact projects (runtime, plugin, generator) from infra related (docs, integration, examples, etc); via prefix (kotlinx-*) or folder (core) - it's not really matter. But still, calling it `core` also sounds strange for me.

In the end, just my 2 cents - may be it's possible to just introduce `generator` project and don't change anything else ? And then later, if needed, reorganise?
I'm not against changes, but I don't see how grouping `runtime` and `generator` makes anything better - may be I'm missing something...

qurbonzoda (approved):
Please make sure to squash all your changes in the PR into one commit before merging (with rebase).

dkrasnoff (approved):

PR #205, opened 2024-04-28
Mark runtime APIs that are unintentionally public with @InternalAPI

qurbonzoda:
A follow-up to https://github.com/Kotlin/kotlinx-benchmark/pull/194


build.gradle
```diff
@@ -68,6 +68,8 @@ apiValidation {
             "integration",
     ]
 
+    nonPublicMarkers += ["kotlinx.benchmark.internal.KotlinxBenchmarkRuntimeInternalApi"]
```
fzhinkin, line 71: Most likely, I'm missing a bit of understanding here, but if, instead of making some API internal or private, we decided to keep it public for a while and only warn its potential users, then it means we do care about binary compatibility and this API should not be filtered from dumps.

And if we don't care about binary compatibility, then we are probably free to make that API internal/private.

runtime/api/kotlinx-benchmark-runtime.klib.api
```diff
@@ -378,78 +99,8 @@ final class kotlinx.benchmark/Blackhole { // kotlinx.benchmark/Blackhole|null[0]
...
-final enum class kotlinx.benchmark/NativeFork : kotlin/Enum<kotlinx.benchmark/NativeFork> { // kotlinx.benchmark/NativeFork|null[0]
-    enum entry PerBenchmark // kotlinx.benchmark/NativeFork.PerBenchmark|null[0]
-    enum entry PerIteration // kotlinx.benchmark/NativeFork.PerIteration|null[0]
-    final fun valueOf(kotlin/String): kotlinx.benchmark/NativeFork // kotlinx.benchmark/NativeFork.valueOf|valueOf#static(kotlin.String){}[0]
-    final fun values(): kotlin/Array<kotlinx.benchmark/NativeFork> // kotlinx.benchmark/NativeFork.values|values#static(){}[0]
-    final val entries // kotlinx.benchmark/NativeFork.entries|#static{}entries[0]
-        final fun <get-entries>(): kotlin.enums/EnumEntries<kotlinx.benchmark/NativeFork> // kotlinx.benchmark/NativeFork.entries.<get-entries>|<get-entries>#static(){}[0]
 }
 // Targets: [js.jsIr, wasmJs]
 final class kotlinx.benchmark/Blackhole : kotlinx.benchmark/CommonBlackhole { // kotlinx.benchmark/Blackhole|null[0]
```
fzhinkin, line 104: It's not directly related to the change, but as soon as `CommonBlackhole` is hidden, there will be no info about `Blackhole`'s members in the dump.

There's a request to overcome that problem on BCV side and dump all members inherited from the private API (https://github.com/Kotlin/binary-compatibility-validator/issues/217), but here, I believe, the better solution would be to introduce a source set descending from `commonMain` and shared between both `jsMain` and `wasmJsMain` source sets and actualize the `Blackhole` there (kotlinx-io use that approach for implementing nodeJs-dependent classes for js and wasmJs: https://github.com/Kotlin/kotlinx-io/blob/c2d52202984f610054f21fd77def42e6f554bb70/build-logic/src/main/kotlin/kotlinx/io/conventions/kotlinx-io-multiplatform.gradle.kts#L84)

runtime/commonMain/src/kotlinx/benchmark/internal/KotlinxBenchmarkPluginInternalApi.kt
```diff
@@ -0,0 +1,19 @@
...
+ * Classes or functions marked with this annotation have `public` visibility
+ * for historical or technical reasons. They are not intended for use by
+ * Kotlinx Benchmark users. They have no behaviour guarantees, and may lack clear
+ * semantics, documentation, and backward compatibility.
+ *
+ * Please remove references to internal API to supported stable references.
+ * If you still require access to these APIs, please raise an issue.
+ */
+@RequiresOptIn(
+    message = "This API is internal to the Kotlinx Benchmark project and will be hidden in a future release.",
```
adam-enko, line 15: This annotation might not necessarily be used on declarations that will be made internal. I suggest changing the message to match [the other opt-in annotation](https://github.com/Kotlin/kotlinx-benchmark/blob/b8e07eb4d70b5f39654c8178d5dd0f8b92fc70c9/plugin/main/src/kotlinx/benchmark/gradle/internal/KotlinxBenchmarkPluginInternalApi.kt#L17). 

```suggestion
    message = "Internal Kotlinx Benchmark API. It may be changed in future releases without notice.",
```

build.gradle
adam-enko: Yes, normally the process is to gradually increment from an opt-in warning, then to an opt-in error, and then remove it. That's not the case here.

@qurbonzoda and I discussed the process regarding the new Gradle plugin opt-in annotation, and we couldn't find any public usage of should-be-internal functions, so we're agreed to skip some steps.

I'm not familiar with the benchmark-runtime API, so maybe the same approach isn't suitable here.

adam-enko (approved):
LGTM (but I'm not a Benchmark user, or familiar with the runtime API, so please take that into account)

build.gradle
fzhinkin: I would expect we either make these functions private/internal and remove them (well, they will be removed automatically) from dumps, or we keep them public and keep them in dumps as well.

qurbonzoda: I see it the following way: 
Currently, we do not provide source or binary compatibility guarantees for the library. We try to make the migration process smooth because we care about our users. 
All APIs annotated with @InternalApi will soon be internal. Thus, going forward, we want to track changes to other public APIs. Again, this doesn’t mean we aren’t going to break compatibility; it's rather to have visibility on what is changing in the public API. Here, those unintentionally public APIs would distort the picture.

qurbonzoda: So, here are two independent efforts going on:
First, the gradual removal of unintentionally public APIs.
Second, keeping track of intentionally public APIs.

qurbonzoda: Does that make sense to you?
If not, I prefer to keep them in dumps for the time being, rather than making them internal right away.

fzhinkin: > Currently, we do not provide source or binary compatibility guarantees for the library.

With that in mind, I don't mind the change. Thanks for elaborating!

fzhinkin (approved):

runtime/api/kotlinx-benchmark-runtime.klib.api
qurbonzoda: Thanks for pointing out. It turns out `CommonBlackhole` can't be properly made `internal`. Because in Kotlin public classes can't inherit from internal classes. I have created a separate PR for sharing `Blackhole` implementation btw Js and WasmJs, so that `CommonBlackhole ` could be dropped later: https://github.com/Kotlin/kotlinx-benchmark/pull/207

PR #207, opened 2024-05-04
Create a separate source set for the shared Blackhole

qurbonzoda:
The CommonBlackhole class is going to be hidden/internalized. Once it is internal, the public Js and WasmJs Blackhole classes won’t be able to inherit from it. This separate source set, shared btw Js and WasmJs, ensures that the targets have a shared implementation, allowing the CommonBlackhole class to be properly internalized or dropped.


runtime/jsWasmJsSharedMain/src/kotlinx/benchmark/CommonBlackhole.kt
```diff
@@ -0,0 +1,48 @@
...
+        arrayOfInt[currentIntPosition] = i
+        currentIntPosition = if (currentIntPosition == MAGIC_SIZE - 1) 0 else currentIntPosition + 1
+    }
+
+    internal fun flushMe() {
+        val sums = arrayOfAny.sumOf { it.hashCode() } + arrayOfInt.sum()
+        println("Consumed blackhole value: $sums")
+    }
+
+    actual fun consume(obj: Any?) = consumeAny(obj)
```
fzhinkin, line 27: These functions are actualized as `inline` on native. Do we need the same here?

qurbonzoda: It does make sense to make the functions inline to eliminate a layer of function calls. However, I wouldn't go further and make the `consumeInt/Any` functions inline. V8 should be able to inline them.

fzhinkin (approved):

adam-enko (approved):

runtime/commonMain/src/kotlinx/benchmark/CommonBlackhole.kt
```diff
@@ -4,6 +4,7 @@ import kotlinx.benchmark.internal.KotlinxBenchmarkRuntimeInternalApi
 
 private const val MAGIC_SIZE: Int = 13
 
+// TODO: Drop this class once it becomes internal
```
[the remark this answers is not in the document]
qurbonzoda: This TODO is part of the `@KotlinxBenchmarkRuntimeInternalApi`. I have mentioned the need to drop the APIs that are unused after they become internal here: https://github.com/Kotlin/kotlinx-benchmark/issues/211

PR #209, opened 2024-05-06
Improve error messages for invalid use of annotations

qurbonzoda:


plugin/main/src/kotlinx/benchmark/gradle/SuiteSourceGenerator.kt
```diff
@@ -139,6 +139,8 @@ class SuiteSourceGenerator(val title: String, val module: ModuleDescriptor, val
             .filterIsInstance<PropertyDescriptor>()
             .filter { it.annotations.any { it.fqName.toString() == paramAnnotationFQN } }
 
+        validateParameterProperties(parameterProperties)
```
fzhinkin, line 142: Currently, neither corresponding annotations have documentation describing restrictions applied to annotated functions/properties/classes, nor does a doc on writing benchmarks mention it.

I believe that we should not only provide a detailed error message to help users in spotting a problem (which you already did!), but also describe how not to write benchmarks, so that they don't do that in the first place.

plugin/main/src/kotlinx/benchmark/gradle/AnnotationsValidator.kt
```diff
@@ -0,0 +1,84 @@
...
+                    "Function `${function.name}` has a parameter of type `$paramType`. ")
+            }
+        } else if (parameters != 0) {
+            error("@Benchmark function can have at most one parameter of type `Blackhole`. " +
+                    "Function `${function.name}` has $parameters parameters.")
+        }
+    }
+}
+
+internal fun validateSetupFunctions(functions: List<FunctionDescriptor>) {
```
fzhinkin, line 33: Note that on JVM, where our annotations are type aliases to JMH ones, one may annotate functions with kotlinx's `Setup`, `Benchmark` and `Teardown`, but then plug JMH-specific parameters into them:

```kotlin
import kotlinx.benchmark.Benchmark
import kotlinx.benchmark.Scope
import kotlinx.benchmark.Setup
import kotlinx.benchmark.State
import org.openjdk.jmh.infra.BenchmarkParams
import org.openjdk.jmh.infra.Control
import org.openjdk.jmh.infra.IterationParams
import org.openjdk.jmh.infra.ThreadParams

@State(Scope.Benchmark) 
public open class SampleBenchmark {
    @Setup 
    public fun setup(bmParams: BenchmarkParams, tParams: ThreadParams, itParams: IterationParams) {
    }

    @Benchmark 
    public fun benchmarkWithControl(ctrl: Control): Boolean = ctrl.stopMeasurement
}
```

All these parameters are a niche feature, but as long as kotlinx-benchmark does not provide alternatives, it would be to allow users to use them.

Here's a more practical example - https://github.com/qurbonzoda/MutableMapEntryBenchmarks/blob/ce674d61b8e9e798fa76c6a926a2129720eac359/src/main/kotlin/PolymorphicMapsBenchmark.kt#L52 ;)

fzhinkin (changes requested):

plugin/main/src/kotlinx/benchmark/gradle/AnnotationsValidator.kt
fzhinkin: Ah, it's only about non-JVM benchmarks. My bad, sorry.

integration/src/main/kotlin/kotlinx/benchmark/integration/AnnotationsSpecifier.kt
```diff
@@ -1,26 +1,109 @@
...
+
+    fun teardown(functionName: String) {
+        functionAnnotations.add(
+            AnnotatedMember(functionName, Annotation("@TearDown"))
+        )
+    }
+
+    fun param(propertyName: String, vararg values: String) {
+        propertyAnnotations.add(
+            AnnotatedMember(propertyName, Annotation("@Param", values.map { "\"$it\"" }))
```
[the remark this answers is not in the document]
qurbonzoda: Done

plugin/main/src/kotlinx/benchmark/gradle/SuiteSourceGenerator.kt
qurbonzoda: I've mentioned the restrictions in the doc on writing benchmarks. Since the annotations have no documentation right now, I believe they should be added in a separate PR.

fzhinkin (approved):

PR #213, opened 2024-05-13
Check NodeJs and D8 versions before adding --experimental-wasm-gc

### LINKED ISSUE #212: WasmJs: don't add experimental-wasm-gc flag for NodeJs >= 22
Starting from version 22, NodeJS is shipped with the V8 that no longer accepts `--experimental-wasm-gc` flag:
```
> Task :wasmJsBenchmark FAILED
Running 'main' benchmarks for 'wasmJs'
~/.gradle/nodejs/node-v22.0.0-darwin-arm64/bin/node: bad option: --experimental-wasm-gc
```

`kotlinx-benchmark` should take V8/NodeJs version into consideration when [adding Wasm options](https://github.com/Kotlin/kotlinx-benchmark/blob/4c7af654240b8efb03ff96895f7846f667bdb7a4/plugin/main/src/kotlinx/benchmark/gradle/JsEngineExecTasks.kt#L68).

fzhinkin:
Closes #212

### DIFF
--- a/plugin/main/src/kotlinx/benchmark/gradle/JsEngineExecTasks.kt
+++ b/plugin/main/src/kotlinx/benchmark/gradle/JsEngineExecTasks.kt
@@ -88,7 +88,7 @@ private fun Project.createNodeJsExec(
     inputFileProperty.set(getExecutableFile(compilation))
     with(nodeArgs) {
         if (compilation.isWasmCompilation) {
-            val addGcArgs = project.extensions.findByType(NodeJsRootExtension::class.java)?.let {
+            val addGcArgs = rootProject.extensions.findByType(NodeJsRootExtension::class.java)?.let {
                 val nodeVersion = VersionNumber.parse(it.nodeVersion)
                 // Starting from version 22, NodeJS is shipped with V8 versions
                 // that no longer accept --experimental-wasm-gc argument.
@@ -116,7 +116,7 @@ private fun Project.createD8Exec(
     description = "Executes benchmark for '${target.name}' with D8"
     inputFileProperty.set(getExecutableFile(compilation))
     if (compilation.isWasmCompilation) {
-        val addGcArgs = project.extensions.findByType(D8RootExtension::class.java)?.let {
+        val addGcArgs = rootProject.extensions.findByType(D8RootExtension::class.java)?.let {
             val d8Version = VersionNumber.parse(it.version)
             // --experimental-wasm-gc flag was removed from V8 starting from ~ 12.3.68
             d8Version < VersionNumber(12, 3, 68, null)
--- a/plugin/main/src/kotlinx/benchmark/gradle/JsEngineExecTasks.kt
+++ b/plugin/main/src/kotlinx/benchmark/gradle/JsEngineExecTasks.kt
@@ -6,8 +6,10 @@ import org.gradle.api.*
 import org.gradle.api.file.RegularFile
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.TaskProvider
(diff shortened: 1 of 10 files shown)


fzhinkin:
As of testing, recent d8/nodeJs versions are only supported starting from `1.9.24`, so we need to either pass Kotlin version to `ProjectBuilder`, or update Kotlin version for the whole project. 

@qurbonzoda, please let me know which version you would prefer, and I'll push tests.

plugin/main/src/kotlinx/benchmark/gradle/JsEngineExecTasks.kt
```diff
@@ -86,7 +88,15 @@ private fun Project.createNodeJsExec(
     inputFileProperty.set(getExecutableFile(compilation))
     with(nodeArgs) {
         if (compilation.isWasmCompilation) {
-            addWasmArguments()
+            val addGcArgs = project.extensions.findByType(NodeJsRootExtension::class.java)?.let {
+                val nodeVersion = VersionNumber.parse(it.nodeVersion)
+                // Starting from version 22, NodeJS is shipped with V8 versions
```
igoriakovlev, line 93: Do we really want to support nodejs version <22? Maybe better just to drop this code as deprecated?

fzhinkin: Sounds good to me. Those who really need an older version can always supply the experimental-gc flag manually (but we should document that workaround, though).

@qurbonzoda WDYT?

lppedd: I'll use an older version of Node for example. I develop VS Code extensions and my Node version is always synced with the Electron one.

fzhinkin: @lppedd, I'm curious: does the Kotlin/Wasm work with it?

lppedd: Yes, I can execute K/WASM tests, and I can load WASM binaries inside of my extensions (e.g., inside of a webview).  
See also https://code.visualstudio.com/blogs/2024/05/08/wasm for what VS Code will support going forward.

Hopefully we'll have a VS Code version based on Electron and Node 22+ by the end of the year, but who knows.

qurbonzoda: Given there are use cases, I would prefer to keep the `--experimental-wasm-gc` argument for users on older Kotlin and/or Node.js versions.

plugin/main/src/kotlinx/benchmark/gradle/JsEngineExecTasks.kt
```diff
@@ -86,7 +88,15 @@ private fun Project.createNodeJsExec(
     inputFileProperty.set(getExecutableFile(compilation))
     with(nodeArgs) {
         if (compilation.isWasmCompilation) {
-            addWasmArguments()
+            val addGcArgs = project.extensions.findByType(NodeJsRootExtension::class.java)?.let {
```
qurbonzoda, line 91: Should probably be `rootProject.extensions.findByType(...)` ?

qurbonzoda:
> As of testing, recent d8/nodeJs versions are only supported starting from 1.9.24, so we need to either pass Kotlin version to ProjectBuilder, or update Kotlin version for the whole project.

I prefer passing the Kotlin version to `ProjectBuilder`.

qurbonzoda (approved):

PR #214, opened 2024-05-22
Use `Classpath` normalization for classpath inputs of `JmhBytecodeGeneratorTask`

4u7:
`@InputFiles` with `@PathSensitive(PathSensitivity.RELATIVE)` makes it impossible to reuse build cache between builds with different build numbers because a jar name includes build number.

https://youtrack.jetbrains.com/issue/KTI-1801/benchmarksmainBenchmarkGenerate-cant-use-build-cache-from-builds-with-different-build-number

### DIFF
--- a/plugin/main/src/kotlinx/benchmark/gradle/JmhBytecodeGeneratorTask.kt
+++ b/plugin/main/src/kotlinx/benchmark/gradle/JmhBytecodeGeneratorTask.kt
@@ -17,12 +17,11 @@ constructor(
     private val workerExecutor: WorkerExecutor
 ) : DefaultTask() {
 
-    @InputFiles
-    @PathSensitive(PathSensitivity.RELATIVE)
+
+    @Classpath
     lateinit var inputClassesDirs: FileCollection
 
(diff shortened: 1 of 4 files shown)


adam-enko:
Good find!

It looks like other Benchmark tasks (WasmSourceGeneratorTask, JsSourceGeneratorTask, NativeSourceGeneratorTask) also use `@InputFiles` instead of `@Classpath` - should they be fixed in this PR as well?

4u7:
>It looks like other Benchmark tasks (WasmSourceGeneratorTask, JsSourceGeneratorTask, NativeSourceGeneratorTask) also use @InputFiles instead of @Classpath - should they be fixed in this PR as well?

I think they should be fixed too, but I don't have a project to test it. In Kotlin only `JmhBytecodeGeneratorTask` is used.

qurbonzoda:
I would prefer consistent path normalization across platforms. If there are any differences, I believe the reasons should be documented.

adam-enko (approved):

PR #216, opened 2024-05-22
Make multiplatform source suite generator be able to process tests in…

### LINKED ISSUE #215: Native benchmark with no package specified fails to compile
Consider the following benchmark that doesn't have a package declaration
```
import kotlinx.benchmark.*
import kotlin.math.*

@State(Scope.Benchmark)
open class RootBenchmark {
    @Benchmark
    open fun mathBenchmark(): Double {
        return log(sqrt(3.0) * cos(3.0), 2.0)
    }
}
```

It fails to compile with:
```
e: build/benchmarks/native/sources/kotlinx/benchmark/generated/%3Croot%3E/RootBenchmark_Descriptor.kt:1:37 Name contains illegal characters: "<>"
e: build/benchmarks/native/sources/kotlinx/benchmark/generated/%3Croot%3E/RootBenchmark_Descriptor.kt:3:8 Unresolved reference: `<root>`
e: build/benchmarks/native/sources/kotlinx/benchmark/generated/%3Croot%3E/RootBenchmark_Descriptor.kt:18:31 Unresolved reference: RootBenchmark
```

SuiteSourceGenerator creates Descriptor with the following code, that is not valid for compiler.
```
package kotlinx.benchmark.generated.`<root>`

import `<root>`.RootBenchmark
```

PavelPunegov:
… the root package.

Fixes #215 issue

### DIFF
--- a/integration/src/test/kotlin/kotlinx/benchmark/integration/SuiteSourceGeneratorTest.kt
+++ b/integration/src/test/kotlin/kotlinx/benchmark/integration/SuiteSourceGeneratorTest.kt
@@ -20,4 +20,25 @@ class SuiteSourceGeneratorTest : GradleTest() {
             }
         }
     }
+
+    @Test
+    fun runBenchmarks() {
+        project("kotlin-multiplatform", true).let { runner ->
+            runner.run(":nativeBenchmarkGenerate")
+
+            runner.generatedDir("native", "RootBenchmark_Descriptor.kt") { descriptorFile ->
+                val text = descriptorFile.readText()
+                println(text)
+                assertFalse(
+                    text.contains("<root>"),
+                    "Generated descriptor contains illegal characters '<root>' in $text"
+                )
+            }
+
+            runner.run(
+                ":compileNativeBenchmarkKotlinNative",
+                ":linkNativeBenchmarkReleaseExecutableNative"
+            )
+        }
+    }
 }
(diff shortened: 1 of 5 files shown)


integration/src/test/kotlin/kotlinx/benchmark/integration/SuiteSourceGeneratorTest.kt
```diff
@@ -20,4 +20,25 @@ class SuiteSourceGeneratorTest : GradleTest() {
             }
         }
     }
+
+    @Test
+    fun runBenchmarks() {
+        project("kotlin-multiplatform", true).let { runner ->
+            runner.run(":nativeBenchmarkGenerate")
```
qurbonzoda, line 27: Could you please also check the generated benchmark sources for `js` and `wasmJs` targets?
It seems like for those targets the `RootBenchmark_Descriptor.kt ` is not generated.

integration/src/test/kotlin/kotlinx/benchmark/integration/SuiteSourceGeneratorTest.kt
```diff
@@ -20,4 +21,38 @@ class SuiteSourceGeneratorTest : GradleTest() {
...
+                val text = descriptorFile.readText()
+                assertFalse(
+                    text.contains("<root>"),
+                    "Generated descriptor contains illegal characters '<root>' in $text"
+                )
+            }
+
+            val capitalizedTarget = target.replaceFirstChar {
+                if (it.isLowerCase()) it.titlecase(Locale.getDefault()) else it.toString()
+            }
```
qurbonzoda, line 54:
```suggestion
            val capitalizedTarget = target.replaceFirstChar { it.uppercaseChar() }
```

`import java.util.*` would also be redundant.

qurbonzoda (approved):

PR #225, opened 2024-06-06
Support debug build configuration for K/N

### LINKED ISSUE #189: Support debug build configuration for K/N
Currently, kotlinx-benchmark implicitly [use](https://github.com/Kotlin/kotlinx-benchmark/blob/231ce03739278a5f0f9805dc2760d568dd7fe64e/plugin/main/src/kotlinx/benchmark/gradle/NativeMultiplatformTasks.kt#L89) release configuration for K/N builds. That's absolutely sane and reasonable default. However, in some cases (like [this](https://youtrack.jetbrains.com/issue/KT-44357)), performance need to be collected and validated for debug builds.

It would be nice to have an option to choose K/N build configuration, with release still being a default.

PavelPunegov:
Adds a build type setting to the NativeBenchmarkTarget to provide the ability to test debug builds. Fixes #189


fzhinkin:
@PavelPunegov, thanks for taking care of it!

It would be nice dropping a few works about the `buildType` in [docs](https://github.com/Kotlin/kotlinx-benchmark/blob/master/docs/configuration-options.md#kotlinnative) (the same way `jmhVersion` described for K/JVM, or `jsBenchmarksExecutor` for K/KS). Otherwise, the change looks good.

PavelPunegov:
@fzhinkin Will do that

PR #227, opened 2024-06-24
Use Gradle Version Catalog for defining dependencies and versions

adam-enko:
Using a Version Catalog helps with sharing dependencies and aligning versions between different projects.

This is particularly useful because Benchmarks uses Included Builds, and Version Catalogs make it easy to make sure the same dependencies are used in each project.

Notes:

- The version of binary-compatibility-validator was `0.15.0-Beta.1` in the root build, but `0.15.0-Beta.2` in the Benchmark Gradle Plugin build. I had to choose the lower version, because `0.15.0-Beta.2` requires a newer version of Gradle.
- The `kotlinx.team.infra` plugin does not have a published Gradle Plugin Marker, so it has to use the legacy plugin application method (buildscript classpath).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,14 +9,14 @@ buildscript {
     }
 
     dependencies {
-        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
-        classpath "kotlinx.team:kotlinx.team.infra:$infra_version"
+        classpath(libs.kotlinx.teamInfraGradlePlugin)
     }
 }
 
 plugins {
     id("base")
-    id("org.jetbrains.kotlinx.binary-compatibility-validator") version "0.15.0-Beta.1"
+    alias(libs.plugins.kotlin.multiplatform) apply false
+    alias(libs.plugins.kotlinx.binaryCompatibilityValidator)
 }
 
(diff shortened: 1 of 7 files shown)


fzhinkin:
@adam-enko 

> I had to choose the lower version, because 0.15.0-Beta.2 requires a newer version of Gradle.

Do you know why the newer Gradle version is required?

adam-enko:
BCV uses a Gradle function:

```
FAILURE: Build failed with an exception.
* What went wrong:
Could not determine the dependencies of task ':kotlinx-benchmark-runtime:apiCheck'.
> Could not create task ':kotlinx-benchmark-runtime:klibApiCheck'.
   > 'void kotlinx.validation.KotlinApiCompareTask.onlyIf(java.lang.String, org.gradle.api.specs.Spec)'
* Try:
* ```

fzhinkin:
It seems like multiplatform plugin is not supported for Gradle `7.4.2` anyway: https://kotlinlang.org/docs/multiplatform-compatibility-guide.html#version-compatibility :)

As of BCV, the issue will be fixed with the next release.

dkrasnoff (approved):

PR #229, opened 2024-07-02
Plugin: Remove Kotlin Reflect dependency

adam-enko:
From what I can tell, the Kotlin Reflect dependency is unnecessary.

To avoid clashes with Gradle's embedded Kotlin, and with other plugins, Gradle plugins should avoid using unnecessary dependencies.


adam-enko:
> There's in fact a single explicit usage of the reflect library in the plugin. :)

Please could you share it here?

PR #231, opened 2024-07-03
Switch expected and actual values in warning for native tasks

OptimumCode:
Hi,

I was a bit confused with the current warning when the native benchmark is skipped because the current OS does not match the expected for that target
E.g.:
> Skipping benchmarks for 'iosArm64' because they cannot be run on current OS: Expected linux_x64, but was ios_arm64

It sounds more like _target **iosArm64** expects **linux_x64** but the current OS is **ios_arm64**_

If you change the expected and actual values it would be more clear. Please, let me know what you think.

If you disagree feel free to close the PR)

### DIFF
--- a/plugin/main/src/kotlinx/benchmark/gradle/NativeMultiplatformTasks.kt
+++ b/plugin/main/src/kotlinx/benchmark/gradle/NativeMultiplatformTasks.kt
@@ -18,7 +18,7 @@ fun Project.processNativeCompilation(target: NativeBenchmarkTarget) {
     val expectedHost = compilation.target.konanTarget
     val actualHost = HostManager.host
     if (expectedHost != actualHost) {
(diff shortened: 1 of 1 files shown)


qurbonzoda (commented):
I agree that the warning could be confusing. Maybe we can simplify the message to: `"Skipping benchmarks for '${target.name}' because they cannot be run on current host '${HostManager.host}'"`?

OptimumCode:
Hi, @qurbonzoda. Thank you for the review. I think we should keep the information about the expected OS in the warning. What do you think about this format for the message?

`"Skipping benchmarks for '${target.name}' because they cannot be run on current host '${actualHost}' (expected host: '${expectedHost}')"`

qurbonzoda:
Great! Please apply the change.

OptimumCode:
Hi, @qurbonzoda. I have applied the suggested changes

PR #232, opened 2024-07-03
Convert Gradle scripts from Groovy DSL to Kotlin DSL

adam-enko:
The Kotlin DSL has many advantages over the Groovy DSL, making it easier to work with.

Some of kotlinx-benchmark's build scripts already use Kotlin DSL, and this PR converts the remaining ones.

I have intentionally tried to keep changes to a minimum, and focused on keeping the same behaviour. There should be no impact on the code or artifacts.

Part of https://youtrack.jetbrains.com/issue/KT-66764


build.gradle.kts
```diff
@@ -19,9 +21,9 @@ plugins {
     alias(libs.plugins.kotlinx.binaryCompatibilityValidator)
 }
 
-apply plugin: 'kotlinx.team.infra'
+apply(plugin = "kotlinx.team.infra")
```
dkrasnoff, line 24: can it be moved to `plugins` block?

dkrasnoff (approved):

build.gradle.kts
```diff
@@ -19,9 +21,9 @@ plugins {
...
 
-infra {
+extensions.configure<InfraExtension> {
     teamcity {
         libraryStagingRepoDescription = project.name
```
dkrasnoff, line 28: I would suggest rewriting this deprecation, as it seems to be a very easy change in the publishing block:
```
        sonatype {
            libraryStagingRepoDescription = project.name
        }
```

build.gradle.kts
adam-enko: No :( There's no [plugin marker artifact](https://docs.gradle.org/8.8/userguide/plugins.html#sec:plugin_markers).

build.gradle.kts
qurbonzoda: I would appreciate the config update. Indeed, it's safer to do it in a separate commit.

qurbonzoda (commented):
Could you please rebase on a fresh master and resolve conflicts?

adam-enko:
> Could you please rebase on a fresh master and resolve conflicts?

Done.


Please could you review this soon? It's difficult trying to keep migrated buildscripts up to date and resolving conflicts, especially because this PR has renamed the files.

build.gradle.kts
```diff
@@ -55,64 +57,68 @@ repositories {
...
-    it.dependsOn(gradle.includedBuild("plugin").task(":apiCheck"))
+    dependsOn(gradle.includedBuild("plugin").task(":apiCheck"))
 }
 
 afterEvaluate {
     gradle.includedBuilds.forEach { included ->
         project(":kotlinx-benchmark-runtime").tasks.named("publishToMavenLocal") { dependsOn(included.task(":publishToMavenLocal")) }
     }
 }
+tasks.register("publishToMavenLocal") {
```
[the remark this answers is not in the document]
adam-enko: I'm not sure, the build config is not easy to decipher.

The code in the `afterEvaluate {}` block makes the `:kotlinx-benchmark-runtime:publishToMavenLocal` task depend on the `:publishToMavenLocal` task of every included build.

This means if any other task depends on `:kotlinx-benchmark-runtime:publishToMavenLocal`, then it will also run the `:publishToMavenLocal` task of every included build. So maybe if you run `./gradlew fooTask` and that requires `:kotlinx-benchmark-runtime:publishToMavenLocal`, and that will trigger the other `:publishToMavenLocal` tasks.

The `tasks.register("publishToMavenLocal)` in the root build doesn't set up the same task dependencies, so they're not the same. It's more useful for being able to quickly run publishing to be able to check it works.

I want to come back and fix the build config, so it's easier to understand and work with. Converting the build scripts to use the Kotlin DSL is just one step of that.

Issue #147, opened 2023-08-21
Update JMH version


fzhinkin:
When running benchmarks on JVM, an old version of JMH is used by default (https://github.com/Kotlin/kotlinx-benchmark/blob/master/gradle.properties#L5).
In addition to numerous improvements made in JMH, there were also some bugs fixed. For example, v1.21 can't properly parse the output of `perfasm` profiler on aarch64 (see https://bugs.openjdk.org/browse/JDK-8252498).

Everyone can specify their own favorite version of JMH manually, but it would be nice to have a more recent version by default.

fzhinkin:
Writing the jmhVersion every single time is pissing me off, so I attempted to bump up the version. Everything works fine as long as there are no tasks specifying older version using `jmhVersion` property. There are no problems if there's only one such a jvm benchmarking target.

However, if there are two or more jvm benchmarking targets and (as it's not prohibited) different jmh versions are set, like here: https://github.com/Kotlin/kotlinx-benchmark/blob/231ce03739278a5f0f9805dc2760d568dd7fe64e/integration/src/test/kotlin/kotlinx/benchmark/integration/SourceSetAsBenchmarkTargetTest.kt#L21
If one such target depend on another,  jmh-dependencies will be updated to the latest version specified (i.e. if there is target with a default jmh version, let's say we updated it to 1.37, and also there's a target with explicitly specified jmh 1.21, then for the latter runtime dependency on jmh will be updated to 1.37; it wouldn't be that bad, but an old JMH version may still be used to generate sources and these sources will be incompatible with a newer JMH version used at compile time).

One solution would be to use strict versions, but that fails short if targets somehow depend on each other (in the example above, where targets are based on source sets, one compilation task will inherit some default dependencies from another and when strict versions are used, that will lead to a conflict).

I'm not sure if the original intent was to allow using different JMH version for different targets within the same project. Assuming it was not, it might be worth reconsidering how this option works and either throwing an exception when there are multiple tasks with different JMH versions, or deprecating target-level option altogether and configuring JMH-version for the whole project only (there's already a `benchmarks_jmh_version` property for that).
Or, of course, we may pursue correct behavior when different tasks use different JMH versions, but personally, I don't see a lot of value there.

@qurbonzoda WDYT?

qurbonzoda:
[...]

Issue #211, opened 2024-05-10
Hide unintentionally public API


qurbonzoda:
Both the `:plugin` and `:runtime` modules have many APIs that were accidentally introduced with a `public` visibility modifier. These APIs are currently annotated with `@KotlinxBenchmarkPluginInternalApi` and `@KotlinxBenchmarkRuntimeInternalApi` in the plugin and runtime, respectively.

The annotations require opt-in with a WARNING level. This level should be advanced to ERROR in a future release, and eventually, the annotated APIs should become `internal`, ceasing to be part of the library's public API.

Once the APIs are made internal, make sure to drop those that are not used internally.

lppedd:
Are you sure the double step is necessary? Given the alpha status I'd not even bother with deprecations and make the internal now. I don't think there are many people using those APIs anyway.

qurbonzoda:
I also believe that not many people use those APIs, if at all. However, I would like to make the migration smoother just in case. The deprecation cycle will be short, unlike in stable libraries.

PR #235, opened 2024-07-04
Use Worker API for Benchmark generators

adam-enko:
Use Worker API to generate benchmark code

- Update the Benchmark JS/Wasm/Native source generation tasks to use the Worker API, so that `kotlin-compiler-embeddable` can be a compileOnly dependency, thus resolving https://youtrack.jetbrains.com/issue/KT-66764
- Create `GenerateJsSourceWorker` and `GenerateWasmSourceWorker` for isolating the JS and Wasm generation code. 
- Updated the Js/Wasm/Native generation tasks to invoke the source generators in a Worker isolated classpath.
- Created a new Configuration for declaring `kotlin-compiler-embeddable`, and pass it to the Js/Wasm/Native generation tasks. 
- Create a BenchmarkDependencies utility for defining all Configurations used in the plugin. (This also lays the groundwork for unifying the JMH version https://github.com/Kotlin/kotlinx-benchmark/issues/147#issuecomment-1980556088) 
- Updated BenchmarksPluginConstants to add `DEFAULT_KOTLIN_COMPILER_VERSION`, as sensible default for kotlin-compiler-embeddable.
- Add opt-in annotation RequiresKotlinCompilerEmbeddable to highlight code that requires `kotlin-compiler-embeddable`
- Some classes were updated to be `abstract`, to follow Gradle best practices for creating [managed objects](https://docs.gradle.org/8.8/userguide/properties_providers.html#managed_types).

I have tested this manually by running the Gradle Plugin tests with Kotlin version set to 2.0. However, Benchmarks is not compatible with KGP 2.0 as a KGP function has changed https://github.com/Kotlin/kotlinx-benchmark/issues/236. If I comment out this changed function, then Benchmarks works as expected.

### DIFF
--- a/plugin/main/src/kotlinx/benchmark/gradle/BenchmarksPlugin.kt
+++ b/plugin/main/src/kotlinx/benchmark/gradle/BenchmarksPlugin.kt
@@ -1,15 +1,22 @@
 package kotlinx.benchmark.gradle
 
+import kotlinx.benchmark.gradle.internal.BenchmarkDependencies
 import kotlinx.benchmark.gradle.internal.BenchmarksPluginConstants
+import kotlinx.benchmark.gradle.internal.BenchmarksPluginConstants.DEFAULT_KOTLIN_COMPILER_VERSION
 import kotlinx.benchmark.gradle.internal.BenchmarksPluginConstants.MIN_SUPPORTED_GRADLE_VERSION
 import kotlinx.benchmark.gradle.internal.KotlinxBenchmarkPluginInternalApi
 import org.gradle.api.*
+import org.gradle.api.provider.*
 import org.gradle.util.GradleVersion
+import javax.inject.Inject
 
 @Suppress("unused")
 abstract class BenchmarksPlugin
 @KotlinxBenchmarkPluginInternalApi
-constructor() : Plugin<Project> {
+@Inject
+constructor(
+    private val providers: ProviderFactory,
+) : Plugin<Project> {
 
     companion object {
         const val PLUGIN_ID = "org.jetbrains.kotlinx.benchmark"
@@ -50,7 +57,7 @@ constructor() : Plugin<Project> {
 
     override fun apply(project: Project) = project.run {
         // DO NOT use properties of an extension immediately, it will not contain any user-specified data
-        val extension = extensions.create(BENCHMARK_EXTENSION_NAME, BenchmarksExtension::class.java, project)
+        val extension = createBenchmarksExtension(project)
 
         if (GradleVersion.current() < GradleVersion.version(MIN_SUPPORTED_GRADLE_VERSION)) {
             logger.error("JetBrains Gradle Benchmarks plugin requires Gradle version $MIN_SUPPORTED_GRADLE_VERSION or higher")
@@ -62,19 +69,24 @@ constructor() : Plugin<Project> {
             if (!getKotlinVersion(kotlinPlugin.pluginVersion).isAtLeast(1, 9, 20)) {
                 logger.error("JetBrains Gradle Benchmarks plugin requires Kotlin version 1.9.20 or higher")
             }
+            extension.kotlinCompilerVersion.set(kotlinPlugin.pluginVersion)
         }
 
-        // Create empty task that will depend on all benchmark building tasks to build all benchmarks in a project
+        // Create a lifecycle task that will depend on all benchmark building tasks to build all benchmarks in a project
         val assembleBenchmarks = task<DefaultTask>(ASSEMBLE_BENCHMARKS_TASKNAME) {
             group = BENCHMARKS_TASK_GROUP
             description = "Generate and build all benchmarks in a project"
         }
 
+        val benchmarkDependencies = BenchmarkDependencies(project, extension)
+
+        configureBenchmarkTaskConventions(project, benchmarkDependencies)
+
         // TODO: Design configuration avoidance
         // I currently don't know how to do it correctly yet, so materialize all tasks after project evaluation.
         afterEvaluate {
             extension.configurations.forEach {
-                // Create empty task that will depend on all benchmark execution tasks to run all benchmarks in a project
+                // Create a lifecycle task that will depend on all benchmark execution tasks to run all benchmarks in a project
                 task<DefaultTask>(it.prefixName(RUN_BENCHMARKS_TASKNAME)) {
                     group = BENCHMARKS_TASK_GROUP
                     description = "Execute all benchmarks in a project"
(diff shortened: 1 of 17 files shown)


plugin/main/src/kotlinx/benchmark/gradle/BenchmarksPlugin.kt
```diff
@@ -92,6 +104,12 @@ constructor() : Plugin<Project> {
         }
     }
 
+    private fun createBenchmarksExtension(project: Project): BenchmarksExtension {
+        return project.extensions.create(BENCHMARK_EXTENSION_NAME, BenchmarksExtension::class.java, project).apply {
+            kotlinCompilerVersion.convention(DEFAULT_KOTLIN_COMPILER_VERSION)
```
fzhinkin, line 109: Why do we need a default version if it will be overridden by the Kotlin version used in a user project?

plugin/main/src/kotlinx/benchmark/gradle/internal/BenchmarkDependencies.kt
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+    internal companion object {
+
+        /** Mark this [Configuration] as one that will be used to declare dependencies. */
+        private fun Configuration.declarable() {
+            isCanBeDeclaredCompat = true
+            isCanBeResolved = false
+            isCanBeConsumed = false
+            isVisible = false
```
fzhinkin, line 56: What's the reason for hiding these configurations? Wouldn't it be helpful for build scripts debugging to keep these configs visible?

plugin/main/src/kotlinx/benchmark/gradle/internal/generator/workers/GenerateJsSourceWorker.kt
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Generates JavaScript benchmarking source code.
+ *
+ * This worker requires `kotlin-compiler-embeddable` and *must* be run in an isolated classpath.
+ *
+ * @see kotlinx.benchmark.gradle.JsSourceGeneratorTask
+ */
+@RequiresKotlinCompilerEmbeddable
+internal abstract class GenerateJsSourceWorker : WorkAction<GenerateJsSourceWorker.Params> {
```
fzhinkin, line 25: This and `GenerateWasmSourceWorker` are almost identical. The only exceptions are `useBenchmarkJs` param and a platform supplied to `SuiteSourceGenerator`.

Perhaps, we can use a single set of classes (worker and parameters) instead?

fzhinkin (changes requested):

plugin/main/src/kotlinx/benchmark/gradle/internal/BenchmarkDependencies.kt
```diff
@@ -0,0 +1,79 @@
...
+internal class BenchmarkDependencies(
+    project: Project,
+    benchmarksExtension: BenchmarksExtension,
+) {
+    private val objects: ObjectFactory = project.objects
+
+    private val benchmarkGenerator: Configuration =
+        project.configurations.create("benchmarkGenerator") {
+            it.description =
+                "Internal Kotlinx Benchmarks Configuration. Contains declared dependencies required for running benchmark generators."
```
fzhinkin, line 23: Maybe `kotlinx-benchmark` instead? I don't think we refer to the plugin as `Kotlinx Benchmarks` anywhere else.

plugin/main/src/kotlinx/benchmark/gradle/BenchmarksPlugin.kt
adam-enko: kotlinx-benchmarks can be used in pure Java projects, where KGP isn't present.

https://github.com/Kotlin/kotlinx-benchmark/blob/v0.4.11/examples/java/build.gradle

plugin/main/src/kotlinx/benchmark/gradle/KlibResolver.kt
```diff
@@ -1,22 +1,29 @@
...
+import org.jetbrains.kotlin.library.resolveSingleFileKlib
+import org.jetbrains.kotlin.library.unresolvedDependencies
+import org.jetbrains.kotlin.storage.StorageManager
+import org.jetbrains.kotlin.util.Logger
 import java.io.*
 import org.jetbrains.kotlin.konan.file.File as KonanFile
 
 internal enum class KlibResolver { JS, Native }
 
+@RequiresKotlinCompilerEmbeddable
```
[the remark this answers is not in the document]
adam-enko: Thanks!

plugin/main/src/kotlinx/benchmark/gradle/internal/BenchmarkDependencies.kt
adam-enko: `isVisible` is confusing, its Javadoc doesn't make sense, it's misnamed, and its default should really be `false`.

If a Configuration is 'visible', then it could automatically be added to the subproject's default outgoing archive, and the producing tasks will always be executed. It should always be set to `false`, unless there's an extremely good reason.

https://github.com/gradle/gradle/issues/6875#issuecomment-1084140718

A better name would probably be `isPrimaryOutput()` and the Javadoc would say "Only enable `isPrimaryOutput()` on outgoing Configurations. The contents will always be shared to consumers by default, unless performing variant aware matching, and they request a specific, valid, and available variant."

fzhinkin: Thank you for the clarification!

plugin/main/src/kotlinx/benchmark/gradle/internal/generator/workers/GenerateJsSourceWorker.kt
fzhinkin: > I would like to resolve this, and simplify the code, but can we agree to do it in later PR in order to keep the changes here focused? I can make an issue.

Sure, fixing it later would be fine

adam-enko: https://github.com/Kotlin/kotlinx-benchmark/issues/240

plugin/main/src/kotlinx/benchmark/gradle/JmhBytecodeGeneratorWorker.kt
```diff
@@ -16,32 +16,43 @@
...
+import org.openjdk.jmh.generators.core.BenchmarkGenerator
+import org.openjdk.jmh.generators.core.FileSystemDestination
+import org.openjdk.jmh.generators.reflection.RFGeneratorSource
+import org.openjdk.jmh.util.FileUtils
+import java.io.File
+import java.net.URL
+import java.net.URLClassLoader
 
 @KotlinxBenchmarkPluginInternalApi
+// TODO Make visibility of JmhBytecodeGeneratorWorker `internal` in version 1.0.
```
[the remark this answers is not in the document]
adam-enko: Thanks for checking, I agree. I updated the TODOs to reference #211

fzhinkin:
Everything looks good to me; only need to revamp kotlinx-benchmark's CI before merging.

dkrasnoff (approved):
Good job!

fzhinkin:
@adam-enko please update the branch with the latest changes from `master` and once the CI is green the PR will be merged

adam-enko:
> @adam-enko please update the branch with the latest changes from `master` and once the CI is green the PR will be merged

Done!

Will the PR be blocked because your last GitHub review is 'requested changes'?

fzhinkin (approved):

PR #239, opened 2024-07-09
Tune some kotlin-team-infra plugin parameters to unlock execution on macOS agents

fzhinkin:
Currently, tasks are not scheduled to macOS agents as agents either don't have JDK16 installed, or are Apple Sillicon-based.
This PR overwrites corresponding tunables to unlock execution on macOS agents.

### DIFF
--- a/.teamcity/utils.kt
+++ b/.teamcity/utils.kt
@@ -14,7 +14,7 @@ const val bintrayToken = "%env.BINTRAY_API_KEY%"
 const val libraryStagingRepoDescription = "kotlinx-benchmark"
 
 val platforms = Platform.values()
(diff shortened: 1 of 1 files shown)


fzhinkin:
https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxBenchmark_Build_All/4699606?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true

ilya-g (approved):
Remind me to remove these requirements from kotlinx-team-infra plugin. We don't need jdk7/8 in projects anymore and there should be only x64 JDKs on agents, so there's no need to use _x64 suffix.

PR #242, opened 2024-07-10
Workaround BCV's incompatibility with Kotlin 2.1

fzhinkin:
The change switches to a BCV built with the latest stable `kotlin-metadata-jvm` that supports 2.1 and also places `kotlin-compiler-embeddable` on the buildscript classpath to let BCV work after https://youtrack.jetbrains.com/issue/KT-61706 integration.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,6 +4,8 @@ buildscript {
     repositories {
         maven { url 'https://maven.pkg.jetbrains.space/kotlin/p/kotlinx/maven' }
         gradlePluginPortal()
+        maven { url 'https://maven.pkg.jetbrains.space/kotlin/p/kotlin/dev' }
+        mavenLocal()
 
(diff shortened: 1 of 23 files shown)


build.gradle
```diff
@@ -18,7 +18,7 @@ buildscript {
 
 plugins {
     id("base")
-    id("org.jetbrains.kotlinx.binary-compatibility-validator") version "0.15.0-Beta.1"
```
adam-enko, line 21: It'd be convenient if the `kotlin-community/dev` branch was updated to include the recent changes from master, because the BCV version can be defined in a single place:

https://github.com/Kotlin/kotlinx-benchmark/blob/4f0dffb59eae072c1c263496ec15bff32811599d/gradle/libs.versions.toml#L4

Would that be possible?

fzhinkin: > Would that be possible?

Well, it's not only possible but also mandatory, I guess. :)

PR #245, opened 2024-07-11
Add Kotlin version override

adam-enko:
Add a dependency constraint in the root project to allow overriding the Kotlin version.

The Kotlin version is not overridden in the Benchmark Gradle plugin - this should use the version of Kotlin embedded into Gradle.

Add a check to ensure that the version is overridden.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,6 +14,9 @@ buildscript {
 
         String kotlinVersion = providers.gradleProperty("kotlin_version").orNull
         if (kotlinVersion != null && !kotlinVersion.isBlank()) {
+            // In addition to overriding the Kotlin version in the Version Catalog,
+            // also enforce the KGP version using a dependency constraint.
+            // Constraints are stricter than Version Catalog.
             constraints {
                 classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion")
             }
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,4 @@
+import org.jetbrains.kotlin.gradle.plugin.KotlinPluginWrapperKt
 import tasks.CheckReadmeTask
 
 buildscript {
@@ -10,6 +11,13 @@ buildscript {
 
     dependencies {
         classpath(libs.kotlinx.teamInfraGradlePlugin)
+
+        String kotlinVersion = providers.gradleProperty("kotlin_version").orNull
+        if (kotlinVersion != null && !kotlinVersion.isBlank()) {
+            constraints {
+                classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion")
+            }
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -10,6 +11,13 @@ buildscript {
 
     dependencies {
         classpath(libs.kotlinx.teamInfraGradlePlugin)
+
+        String kotlinVersion = providers.gradleProperty("kotlin_version").orNull
+        if (kotlinVersion != null && !kotlinVersion.isBlank()) {
+            constraints {
+                classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion")
```
fzhinkin, line 18: Should we do something like that instead: https://docs.gradle.org/current/userguide/platforms.html#sec:overwriting-catalog-versions?
To ensure that the proper version will be propagated everywhere (here, for example: https://github.com/Kotlin/kotlinx-benchmark/blob/0764551f1dde18955f986fbf1dd581449202a23f/integration/build.gradle.kts#L27)

build.gradle
```diff
@@ -54,8 +62,15 @@ afterEvaluate {
 }
 //endregion
 
+String kgpVersion = KotlinPluginWrapperKt.getKotlinPluginVersion(project)
+logger.info("Using Kotlin ${KotlinPluginWrapperKt.getKotlinPluginVersion(it)}")
```
fzhinkin, line 66: I'm getting `> Could not get unknown property 'it' for root project 'kotlinx-benchmark' of type org.gradle.api.Project.` locally.

adam-enko: oops, bad copy/paste. I pushed a fix.

build.gradle
adam-enko: Very good point. Probably yes, it makes sense to override the version to make sure it's aligned. I wanted to avoid it because it will affect the KGP used to build the Gradle plugin, but actually, now that I think about it, the Gradle plugin is an included build and so it shouldn't be affected...

I'll try fixing it.

adam-enko: Done, version catalog is overridden in `settings.gradle`, please take a look.

gradle/libs.versions.toml
```diff
@@ -1,5 +1,6 @@
 [versions]
 
+# Note: Kotlin version can be overridden by passing `-Pkotlin_dev=<version>`
```
fzhinkin, line 3:
```suggestion
# Note: Kotlin version can be overridden by passing `-Pkotlin_version=<version>`
```
Right?

build.gradle
fzhinkin: Thanks! Do we still need this particular change with the `constraint` if now the version in the catalog will be overridden by a `kotlin_version` value?

gradle/libs.versions.toml
adam-enko: yes, thanks!

build.gradle
adam-enko: Hmmm practically we could remove the constraint, but Gradle dependency resolution can be  is weird and confusing and I think it would be easy to accidentally modify the resolved version. The assertions would help catch this, but I'd feel more confident keeping the constraint.

fzhinkin: Ok. I don't have a particular opinion about the assertion, perhaps @dkrasnoff could add something about it.

Otherwise, the change looks good to me.

fzhinkin (approved):

build.gradle
dkrasnoff: To me the only place looks more convenient to support.
If we have two places we should somewhere comment on why we have these two places, just help "future us" to remember :)

adam-enko: I'd be happy to document it, but I'm not sure where...

Maybe CONTRIBUTING.md? But it doesn't exist yet, despite a link in the README...

https://github.com/Kotlin/kotlinx-benchmark/blob/0764551f1dde18955f986fbf1dd581449202a23f/README.md?plain=1#L435

dkrasnoff: What about leaving a comment right here, where you add constraints with an explanation?

adam-enko: Good idea, done. I also added a brief explanation to the Version Catalog override.

dkrasnoff:
Changes look good for me.
Good job 🆒 👍

PR #249, opened 2024-07-23
Introduce a workaround for apiCheck similar to apiDump

### LINKED ISSUE #248: apiCheck isn't called for the Gradle plugin
https://github.com/Kotlin/kotlinx-benchmark/commit/4d7799973f789855a52b0d8e8e9567c01cddfe29 introduced BCV for the Gradle plugin.
Also, it introduced the workaround
```
tasks.register("apiDump") {
    it.dependsOn(gradle.includedBuild("plugin").task(":apiDump"))
}
```
however, currently, `./gradlew apiCheck` does not check the API dump of the plugin, and it is indeed outdated
Perhaps, there should be added a similar workaround for `apiCheck`.

ALikhachev:
Also, update the dump to reflect the actual state of the Gradle plugin. Reorders entries in the dumps: https://github.com/Kotlin/binary-compatibility-validator/pull/225, https://github.com/Kotlin/binary-compatibility-validator/issues/196 Removes internal constants from the dump: https://github.com/Kotlin/binary-compatibility-validator/issues/90 Fixes #248

### DIFF
--- a/plugin/api/kotlinx-benchmark-plugin.api
+++ b/plugin/api/kotlinx-benchmark-plugin.api
@@ -34,11 +34,12 @@ public class kotlinx/benchmark/gradle/BenchmarkTarget {
 	public final fun setWorkingDir (Ljava/lang/String;)V
 }
 
-public class kotlinx/benchmark/gradle/BenchmarksExtension {
+public abstract class kotlinx/benchmark/gradle/BenchmarksExtension {
 	public final fun configurations (Lgroovy/lang/Closure;)Lorg/gradle/api/NamedDomainObjectContainer;
 	public final fun getBenchsDescriptionDir ()Ljava/lang/String;
 	public final fun getBuildDir ()Ljava/lang/String;
 	public final fun getConfigurations ()Lorg/gradle/api/NamedDomainObjectContainer;
+	public abstract fun getKotlinCompilerVersion ()Lorg/gradle/api/provider/Property;
 	public final fun getProject ()Lorg/gradle/api/Project;
 	public final fun getReportsDir ()Ljava/lang/String;
 	public final fun getTargets ()Lorg/gradle/api/NamedDomainObjectContainer;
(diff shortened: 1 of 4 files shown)


adam-enko (approved):

plugin/api/kotlinx-benchmark-plugin.api
```diff
@@ -34,11 +34,12 @@ public class kotlinx/benchmark/gradle/BenchmarkTarget {
 	public final fun setWorkingDir (Ljava/lang/String;)V
 }
 
-public class kotlinx/benchmark/gradle/BenchmarksExtension {
+public abstract class kotlinx/benchmark/gradle/BenchmarksExtension {
 	public final fun configurations (Lgroovy/lang/Closure;)Lorg/gradle/api/NamedDomainObjectContainer;
 	public final fun getBenchsDescriptionDir ()Ljava/lang/String;
 	public final fun getBuildDir ()Ljava/lang/String;
 	public final fun getConfigurations ()Lorg/gradle/api/NamedDomainObjectContainer;
+	public abstract fun getKotlinCompilerVersion ()Lorg/gradle/api/provider/Property;
```
qurbonzoda, line 42: @adam-enko Could you please clarify if this property is intended to be configured by kotlinx-benchmark users?

adam-enko: Typically users shouldn't need to configure it, but I left it as public in case they need to update the Benchmarks plugin but keep using an older `kotlin-compiler-embeddable`.

qurbonzoda: > In case they need to update the Benchmarks plugin but keep using an older `kotlin-compiler-embeddable`.

Do you mean in other parts of there benchmark module?

adam-enko: I mean in the subproject that has the kotlinx-benchmark Gradle plugin.

PR #250, opened 2024-08-09
Implement support for Android target

kx412764776:


examples/kotlin-multiplatform/build.gradle
```diff
@@ -1,17 +1,55 @@
...
+//    > Could not find com.android.tools.lint:lint-gradle:31.2.2.
+repositories {
+    google()
+}
+
 kotlin {
+    androidTarget {
+        // Android target does not have any compilations
+        println("android compilations: ${compilations.size()}")
+    }
```
fzhinkin, line 52: JFYI: when `implementation project(":kotlinx-benchmark-runtime")` is replaced with a dependency on actual artifact (for instance, published to a local maven repo), `:compileReleaseKotlinAndroid` fails due to unresolved `kotlinx.benchnark` symbols:
```
e: file:///Users/filipp.zhinkin/Development/kx-benchmark-android/examples/kotlin-multiplatform/src/androidMain/kotlin/AndroidTestBenchmark.kt:6:7 This class does not have a constructor
e: file:///Users/filipp.zhinkin/Development/kx-benchmark-android/examples/kotlin-multiplatform/src/androidMain/kotlin/AndroidTestBenchmark.kt:6:14 Unresolved reference: Benchmark
e: file:///Users/filipp.zhinkin/Development/kx-benchmark-android/examples/kotlin-multiplatform/src/androidMain/kotlin/AndroidTestBenchmark.kt:7:8 This class does not have a constructor
e: file:///Users/filipp.zhinkin/Development/kx-benchmark-android/examples/kotlin-multiplatform/src/androidMain/kotlin/AndroidTestBenchmark.kt:8:13 This class does not have a constructor
```

It does not make much sense to redefine a dependency for this sample project, but the same issue is also reproduced on `kotlinx-io` when I tried to apply kx-benchmark w/ android support.

plugin/main/src/kotlinx/benchmark/gradle/AndroidMultiplatformTasks.kt
```diff
@@ -0,0 +1,219 @@
...
+
+                generateBenchmarkSourceFiles(targetDir, classDescriptors)
+            }
+        }
+    }
+}
+
+private fun detectAndroidDevice() {
+    println("Detect running Android devices...")
+    val devices = ProcessBuilder("adb", "devices")
```
fzhinkin, line 107: `adb` may not be on a `PATH`.
In that case it makes sense to check `sdk.dir` property, `ANDROID_HOME` env variable, or , what seems to be a proper way,  [ask Android plugin for an ADB path](https://developer.android.com/reference/tools/gradle-api/8.3/null/com/android/build/api/dsl/SdkComponents#getAdb()).

plugin/main/resources/GeneratedAndroidProject/microbenchmark/build.gradle.kts
```diff
@@ -0,0 +1,43 @@
+@Suppress("DSL_SCOPE_VIOLATION") // TODO: Remove once KTIJ-19369 is fixed
+plugins {
+    alias(libs.plugins.android.library)
+    alias(libs.plugins.jetbrains.kotlin.android)
+}
+
+android {
+    namespace = "kotlinx.benchmark.android.microbenchmark"
+    compileSdk = 34
```
fzhinkin, line 9: When generating a project, compile SDK and runtime SDK versions as well as other properties, like defaultConfig and compileOptions, should be overridden by values inferred from a project the plugin was applied to.

plugin/main/resources/GeneratedAndroidProject/microbenchmark/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,2 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"/>
```
fzhinkin, line 2: In future, it would be nice to provide a way to supply a user-defined config (it may be required to setup profiling, for example).

PR #252, opened 2024-08-23
Fix Native Benchmark Compilations triggering compile tasks on Gradle sync

sellmair:
... by using `associateWith` instead of manually wiring up compile outputs back to the `implementation` dependency of the benchmark compilation.

Tested using the `kotlinx.io` project. 
**Before**:
before

**After**:
image

### DIFF
--- a/plugin/main/src/kotlinx/benchmark/gradle/NativeMultiplatformTasks.kt
+++ b/plugin/main/src/kotlinx/benchmark/gradle/NativeMultiplatformTasks.kt
@@ -74,14 +74,8 @@ private fun Project.createNativeBenchmarkCompileTask(target: NativeBenchmarkTarg
         sourceSet.resources.setSrcDirs(files())
         sourceSet.kotlin.setSrcDirs(files("$benchmarkBuildDir/sources"))
 
-        sourceSet.dependencies {
-            implementation(compilation.output.allOutputs)
-        }
-        project.configurations.let {
-            it.getByName(sourceSet.implementationConfigurationName).extendsFrom(
-                it.getByName(compilation.compileDependencyConfigurationName)
-            )
-        }
+        associateWith(compilation)
+
 
         // TODO: check if there are other ways to set compiler options.
         this.kotlinOptions.freeCompilerArgs = compilation.kotlinOptions.freeCompilerArgs


fzhinkin:
@sellmair thanks for taking care of the problem!

plugin/main/src/kotlinx/benchmark/gradle/NativeMultiplatformTasks.kt
```diff
@@ -74,14 +74,8 @@ private fun Project.createNativeBenchmarkCompileTask(target: NativeBenchmarkTarg
         sourceSet.resources.setSrcDirs(files())
         sourceSet.kotlin.setSrcDirs(files("$benchmarkBuildDir/sources"))
 
-        sourceSet.dependencies {
-            implementation(compilation.output.allOutputs)
```
fzhinkin, line 78: @sellmair could you please clarify why it solves the issue?

I'm looking into `KotlinNativeCompilationAssociator`, and its implementation looks almost the same: https://github.com/JetBrains/kotlin/blob/ad8cea5f0c154598a45bc8f720c07d8de9fe19f1/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/plugin/mpp/compilationImpl/KotlinCompilationAssociator.kt#L69.

sellmair: Of course! You see the main difference is that kotlinx.benchmark used the `implementation` configuration to attach the dependencies. This will be resolved by the IDE for dependencies. However, the `compileDependencyFiles` or `compileDependencyConfiguration` is at the very end of the dependency-resolution pipeline and will not cause the IDE resolve this.

sellmair: So in short: `implementation` is "for users declaring dependencies"
'compileDependencyFiles/compileDependencyConfiguration' is the actual 'lower level' container for the compile classpath

fzhinkin: Thanks!

fzhinkin (approved):
LGTM

qurbonzoda (commented):
Do the [JS](https://github.com/Kotlin/kotlinx-benchmark/blob/0af2a2b5c4081cca61a1e88753bbd97edcd328cf/plugin/main/src/kotlinx/benchmark/gradle/JsMultiplatformTasks.kt#L40) and [WasmJs](https://github.com/Kotlin/kotlinx-benchmark/blob/4d7799973f789855a52b0d8e8e9567c01cddfe29/plugin/main/src/kotlinx/benchmark/gradle/WasmMultiplatformTasks.kt#L39) benchmark compilations also benefit from the same change?

sellmair:
> Do the [JS](https://github.com/Kotlin/kotlinx-benchmark/blob/0af2a2b5c4081cca61a1e88753bbd97edcd328cf/plugin/main/src/kotlinx/benchmark/gradle/JsMultiplatformTasks.kt#L40) and [WasmJs](https://github.com/Kotlin/kotlinx-benchmark/blob/4d7799973f789855a52b0d8e8e9567c01cddfe29/plugin/main/src/kotlinx/benchmark/gradle/WasmMultiplatformTasks.kt#L39) benchmark compilations also benefit from the same change?

I am not aware of those targets suffering the same issue 🤔

Issue #236, opened 2024-07-04
NoSuchMethodError in project that uses Kotlinx Benchmark and KGP 2.0


adam-enko:
kotlinx-benchmark uses a utility to create a new D8Exec task.

https://github.com/Kotlin/kotlinx-benchmark/blob/1aefc62c585362e2dca2f7b695519e4d7c5052cf/plugin/main/src/kotlinx/benchmark/gradle/JsEngineExecTasks.kt#L108-L113

However, the signature of this method changed in KGP 2.0

https://github.com/JetBrains/kotlin/blob/9589486e67c5867b41ed6c58c9614aff39633fe3/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/targets/js/d8/D8Exec.kt#L54

`compilation: KotlinJsCompilation` changed to `compilation: KotlinJsIrCompilation`

This causes a failure during Gradle project configuration.

```
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'junit2932190119581877902'.
> Failed to notify project evaluation listener.
   > 'org.gradle.api.tasks.TaskProvider org.jetbrains.kotlin.gradle.targets.js.d8.D8Exec$Companion.create(org.jetbrains.kotlin.gradle.plugin.mpp.KotlinJsCompilation, java.lang.String, kotlin.jvm.functions.Function1)'

* Try:
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Exception is:
org.gradle.api.ProjectConfigurationException: A problem occurred configuring root project 'junit2932190119581877902'
[...]
Caused by: java.lang.NoSuchMethodError: 'org.gradle.api.tasks.TaskProvider org.jetbrains.kotlin.gradle.targets.js.d8.D8Exec$Companion.create(org.jetbrains.kotlin.gradle.plugin.mpp.KotlinJsCompilation, java.lang.String, kotlin.jvm.functions.Function1)'
	at kotlinx.benchmark.gradle.JsEngineExecTasksKt.createD8Exec(JsEngineExecTasks.kt:113)
```

fzhinkin:
As of now, everything works fine w/ latest KGP.

PR #255, opened 2024-09-02
Bump Kotlin version to 2.0.20

qurbonzoda:
* Bump Kotlin version to 2.0.20
* Drop obsolete tests
* Remove IR designation for JS target declarations
* Remove redundant dependsOn edges in runtime module
* Bump the minimum supported Kotlin version to 2.0.0 and verify it through tests


integration/src/test/kotlin/kotlinx/benchmark/integration/WasmGcOptionsTest.kt
```diff
@@ -2,24 +2,18 @@ package kotlinx.benchmark.integration
...
 
+// TODO: Remove this test and the associated code that adds the "--experimental-wasm-gc" flag once
+//  the minimum supported Kotlin version no longer supports D8 < 12.3.68.
 class WasmGcOptionsTest : GradleTest() {
-    @Test
-    fun nodeJs() {
-        // The test uses Kotlin 1.9.24 as previous versions
-        // would append the --experimental-wasm-gc flag causing run failures.
-        val runner = project(
-            "wasm-gc-non-experimental/wasm-nodejs", true, kotlinVersion = "1.9.24"
```
qurbonzoda, line 11: Since the supported Kotlin versions are not compatible with NodeJs versions earlier than 22.0.0, I am not sure if this use case is still viable: https://github.com/Kotlin/kotlinx-benchmark/pull/213#discussion_r1598490631.

If it is, the code that adds the `--experimental-wasm-gc` flag to `NodeJsExec` should be preserved.

However, this test does not seem to make much sense in any case, since the supported Kotlin versions use NodeJs 22.0.0 or newer, and setting the `nodeVersion` to an earlier version will cause the test project to fail.

README.md
```diff
@@ -10,7 +10,7 @@
 `kotlinx-benchmark` is a toolkit for running benchmarks for multiplatform code written in Kotlin.
 It is designed to work with Kotlin/JVM, Kotlin/JS, Kotlin/Native, and Kotlin/Wasm (experimental) targets.
 
-To get started, ensure you're using Kotlin 1.9.20 or newer and Gradle 7.4 or newer.
+To get started, ensure you're using Kotlin 2.0.0 or newer and Gradle 7.4 or newer.
```
qurbonzoda, line 13: Min supported version was bumped chiefly because of https://github.com/Kotlin/kotlinx-benchmark/issues/236

examples/kotlin/build.gradle.kts
adam-enko, line 1: Unrelated question: Now that the `kotlin` example has been converted from Groovy to Kotlin DSL, is there any difference between it and `kotlin-kts`? Maybe one should be removed?

examples/kotlin-kts/build.gradle.kts
adam-enko, line 1: `kotlinOptions {}` is deprecated.

```kotlin
tasks.withType<KotlinCompile> {
    kotlinOptions {
        jvmTarget = "1.8"
    }
}
```

Please replace with `compilerOptions {}`

```kotlin
kotlin {
    compilerOptions {
        jvmTarget = JvmTarget.JVM_1_8
    }
}
```

examples/kotlin-kts/build.gradle.kts
```diff
@@ -8,7 +8,7 @@ import org.jetbrains.kotlin.gradle.tasks.*
 plugins {
     java
     kotlin("jvm")
```
adam-enko, line 10: Applying both KGP and a Java plugin causes issues and is deprecated, so remove Java and the `tasks.withType<JavaCompile> {` config below.

https://kotlinlang.org/docs/whatsnew2020.html#deprecated-compatibility-with-kotlin-multiplatform-gradle-plugin-and-gradle-java-plugins

adam-enko (changes requested):
There are some small things to tidy up.

adam-enko:
Please add the `.kotlin/` directory to `.gitignore`

examples/kotlin/build.gradle.kts
qurbonzoda: Yes, one should be removed.

integration/src/test/kotlin/kotlinx/benchmark/integration/WasmGcOptionsTest.kt
qurbonzoda: @adam-enko Could you please help me understand if the use case mentioned is still viable with Kotlin 2.0.20?

examples/kotlin/build.gradle.kts
qurbonzoda: Sorry for the misleading information. After a more thorough look, it turns out that the two example projects also demonstrate different benchmark project configurations. To better reflect this, I have renamed the example projects in a separate PR: https://github.com/Kotlin/kotlinx-benchmark/pull/257

examples/kotlin-kts/build.gradle.kts
qurbonzoda: I have addressed this in the followup PR: https://github.com/Kotlin/kotlinx-benchmark/pull/257

examples/kotlin-kts/build.gradle.kts
qurbonzoda: IIUC, simultaneous usage of KMP and Java plugin is deprecated. Whereas, here kotlin-jvm plugin is used. Please correct me if I got it wrong.

Anyways, the `java` plugin is redundant here and is removed in a followup PR: https://github.com/Kotlin/kotlinx-benchmark/pull/257

adam-enko: You're right! My mistake.

integration/src/test/kotlin/kotlinx/benchmark/integration/WasmGcOptionsTest.kt
fzhinkin: I tried to build a 2.0.20-based project with NodeJS `21.0.0-v8-canary-*` and after a few a options added, everything continue working the same way it does with v22.

I don't know why one would do that in their projects, but it works.

plugin/build.gradle
adam-enko, line 1: Please can you review https://github.com/Kotlin/kotlinx-benchmark/pull/232 before merging this PR? It'd be nice to avoid re-basing it again.

plugin/main/src/kotlinx/benchmark/gradle/NativeMultiplatformTasks.kt
```diff
@@ -123,11 +125,11 @@ fun Project.createNativeBenchmarkExecTask(
 
         val binary =
             benchmarkCompilation.target.binaries.getExecutable(benchmarkCompilation.name, target.buildType)
-        val linkTask = binary.linkTask
+        val linkTask = binary.linkTaskProvider
 
         dependsOn(linkTask)
 
-        val executableFile = linkTask.outputFile.get()
+        val executableFile = linkTask.get().outputFile.get()
```
adam-enko, line 132: `.get()` will eagerly configure the task and the value. 

It'd probably work instead like this:

```kotlin
onlyIf { linkTask.get().outputFile.get().exists() }
```

qurbonzoda: Could you please help me with the other use of `executableFile` below?

Should it be `linkTask.configure { this.executable.set(it.outputFile) }` ?
This would also require changing `NativeBenchmarkExec.executable` type to `Property<File>`.

adam-enko: Ah I see... yes, `NativeBenchmarkExec.executable` should be updated to be compatible with the Provider API. Do you want do that now, or in another PR?

(BTW `RegularFileProperty` or `DirectoryProperty` is preferred over `Provider<File>`.)

adam-enko (approved):

plugin/main/src/kotlinx/benchmark/gradle/NativeMultiplatformTasks.kt
qurbonzoda: I will do it in a separate PR.

Issue #176, opened 2023-12-27
Plugin's tasks ignore configured java toolchain


fzhinkin:
`kotlinx-benchmarks` tasks responsible for compiling and running benchmarks on JVM ignore a toolchain configured for a project and instead use the same JDK as Gradle do.

That to leads to compilation failures when a project is configured to use a toolchain that is newer than the java Gradle was launched with.

For example, a project's toolchain is JDK 21, but Gradle was started with JDK 17:
```
Execution failed for task ':kotlinx-io-benchmarks:jvmBenchmarkGenerate'.
> There was a failure while executing work items
   > A failure occurred while executing kotlinx.benchmark.gradle.JmhBytecodeGeneratorWorker
      > kotlinx/io/benchmarks/ShortBenchmark has been compiled by a more recent version of the Java Runtime (class file version 65.0), this version of the Java Runtime only recognizes class file versions up to 61.0
```

It would be nice to align kotlinx-benchmarks' tasks behavior with Kotlin Gradle plugins and use JDK configured using toolchains.

fzhinkin:
Related feature that would be nice to support (but it probably deserves a separate issue): allow configuring different toolchains for different benchmarks configurations, so that a user could execute the same set of benchmarks with different JDK versions. Something like:
```
configurations {
    create("jdk17") {
        mode = "avgt"
        outputTimeUnit = "ns"
        include("MyPreciousBenchmark")
        jvmToolchain {
            languageVersion.set(JavaLanguageVersion.of(17))
        }
    }
    create("jdk21") {
        mode = "avgt"
        outputTimeUnit = "ns"
        include("MyPreciousBenchmark")
        jvmToolchain {
            languageVersion.set(JavaLanguageVersion.of(21))
        }
    }
}
```

fzhinkin:
Filed https://github.com/Kotlin/kotlinx-benchmark/issues/187 to track https://github.com/Kotlin/kotlinx-benchmark/issues/176#issuecomment-1870419229

PR #257, opened 2024-09-05
Use jvmToolchain(8) for the plugin and runtime modules

qurbonzoda:
Otherwise, if the project using kotlinx-benchmark is configured with an older jvmToolchain than the Java with which Gradle was launched to assemble plugin or runtime, the following error occurs:
```
Execution failed for task ':jvmBenchmarkGenerate'.
> There was a failure while executing work items
  > A failure occurred while executing kotlinx.benchmark.gradle.JmhBytecodeGeneratorWorker
    > kotlinx/benchmark/gradle/JmhBytecodeGeneratorWorkParameters has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```
Similar issue: https://github.com/Kotlin/kotlinx-benchmark/issues/176

Additionally:
* Remove redundant `java` plugin application in example projects
* Rename example projects `kotlin-kts` -> `kotlin-jvm` and `kotlin` -> `kotlin-jvm-separate-benchmark-source-set` to better reflect what they meant to demonstrate


qurbonzoda:
The base branch will be changed to `master` once `qurbonzoda/bump-kotlin-version` gets merged.

fzhinkin (approved):

examples/kotlin-jvm/build.gradle.kts
```diff
@@ -25,16 +23,8 @@ dependencies {
...
-}
-
-
-tasks.withType<KotlinCompile> {
-    kotlinOptions {
-        jvmTarget = "1.8"
-    }
+kotlin {
+    jvmToolchain(8)
 }
```
adam-enko, line 28: Maybe instead of using Java 8, we could use Java 21 to build and set the release flag to make the JVM classes target Java 8?

Also, use a Java 8 to launch and run the tests.

qurbonzoda: > Maybe instead of using Java 8, we could use Java 21 to build and set the release flag to make the JVM classes target Java 8?

Do you mean in the example project?

qurbonzoda: > Also, use a Java 8 to launch and run the tests.

Done, PTAL

adam-enko (approved):
It'd be nice to use JVM 21 and `-release 8`, but that can be done later, and this PR is a good incremental improvement.

It's good to see kxb improving!

ilya-g (approved):

PR #261, opened 2024-09-12
Use fixated Kotlin compiler for compiling the Gradle plugin.

ALikhachev:
Relates to KT-60521

### DIFF
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -1,5 +1,7 @@
 import org.jetbrains.kotlin.gradle.dsl.KotlinVersion
 import kotlinx.team.infra.InfraExtension
+import org.jetbrains.kotlin.buildtools.api.ExperimentalBuildToolsApi
+import org.jetbrains.kotlin.gradle.ExperimentalKotlinGradlePluginApi
 
 buildscript {
     repositories {
@@ -83,14 +85,23 @@ sourceSets {
     }
 }
 
+kotlin {
+    @OptIn(ExperimentalBuildToolsApi::class, ExperimentalKotlinGradlePluginApi::class)
+    compilerVersion = "2.0.20" // Kotlin 2.1 removes support for the used language version / api version: KT-60521
+}
+
 tasks.compileKotlin {
     compilerOptions {
         optIn.addAll(
                 "kotlinx.benchmark.gradle.internal.KotlinxBenchmarkPluginInternalApi",
                 "kotlin.RequiresOptIn",
         )
-        @Suppress("DEPRECATION")
-        apiVersion = KotlinVersion.KOTLIN_1_4 // the version of Kotlin embedded in Gradle
+        /*
+         * Those versions are configured according to https://docs.gradle.org/current/userguide/compatibility.html
+         * and the Kotlin compiler compatibility policy stating that Kotlin 1.4 is compatible with 1.5 binaries
+         */
+        languageVersion = KotlinVersion.KOTLIN_1_5
+        apiVersion = KotlinVersion.KOTLIN_1_4
     }
 }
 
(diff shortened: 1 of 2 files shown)


plugin/build.gradle.kts
```diff
@@ -83,14 +85,26 @@ sourceSets {
...
-        apiVersion = KotlinVersion.KOTLIN_1_4 // the version of Kotlin embedded in Gradle
+        /**
+         * Those versions are configured according to https://docs.gradle.org/current/userguide/compatibility.html
+         * and the Kotlin compiler compatibility policy stating that Kotlin 1.4 is compatible with 1.5 binaries
+         */
+        @Suppress("DEPRECATION", "DEPRECATION_ERROR")
+        run {
+            languageVersion = KotlinVersion.KOTLIN_1_5
+            apiVersion = KotlinVersion.KOTLIN_1_4
+        }
```
qurbonzoda, line 107: Could you please elaborate on this part a little more?

If I understand correctly, Kotlin compiler 2.1 will drop support for `language/apiVersion` 1.4 and 1.5. 
Also, KGP 2.1 supports Gradle 7.6.3+, which, according to the [Gradle compatibility guide](https://docs.gradle.org/current/userguide/compatibility.html), comes with Kotlin 1.7.10 and only supports Kotlin language features available in 1.4.
Therefore, to ensure the kx-benchmark plugin continues supporting Gradle 7.6.3+ projects, we use compiler version 2.0.20 for compiling the plugin. This allows us to set `languageVersion = 1.4` (or 1.5, which is compatible with 1.4 binaries).

Please correct me if I got it wrong.

My other question is: What version should apiVersion indicate? Is it the version of the embedded Kotlin in the minimal supported Gradle? For example, for Gradle 7.6.3, should it be 1.7?

plugin/build.gradle.kts
```diff
@@ -83,14 +85,26 @@ sourceSets {
...
     compilerOptions {
         optIn.addAll(
                 "kotlinx.benchmark.gradle.internal.KotlinxBenchmarkPluginInternalApi",
                 "kotlin.RequiresOptIn",
         )
-        @Suppress("DEPRECATION")
-        apiVersion = KotlinVersion.KOTLIN_1_4 // the version of Kotlin embedded in Gradle
+        /**
+         * Those versions are configured according to https://docs.gradle.org/current/userguide/compatibility.html
+         * and the Kotlin compiler compatibility policy stating that Kotlin 1.4 is compatible with 1.5 binaries
```
qurbonzoda, line 101: Is this information public?

plugin/build.gradle.kts
ALikhachev: That's right. 

> My other question is: What version should apiVersion indicate? Is it the version of the embedded Kotlin in the minimal supported Gradle? For example, for Gradle 7.6.3, should it be 1.7?

`apiVersion` affects available API from Kotlin stdlib. Thus, you're right again. It should be equal to the version of Kotlin stdlib provided by Gradle at runtime.

plugin/build.gradle.kts
ALikhachev: Yes. 
https://kotlinlang.org/docs/kotlin-evolution-principles.html#evolving-the-binary-format

> Preferably (but we can't guarantee it), the binary format is mostly forwards compatible with the next language release, but not later ones (in the cases when new features are not used, for example, 1.9 can understand most binaries from 2.0, but not 2.1).

qurbonzoda: Thanks for the clarification!

plugin/build.gradle.kts
qurbonzoda: Thanks for the clarification!

PR #264, opened 2024-10-02
Override Kotlin apiVersion and languageVersion compiler options …

qurbonzoda:
…with Gradle properties.

This is needed for building the library as a part of TeamCity aggregate builds.

### DIFF
--- a/buildSrc/src/main/kotlin/KotlinCommunity.kt
+++ b/buildSrc/src/main/kotlin/KotlinCommunity.kt
@@ -3,7 +3,6 @@
 import org.gradle.api.Project
 import org.gradle.api.artifacts.dsl.*
 import java.net.*
-import java.util.logging.Logger
 
 /*
  * Functions in this file are responsible for configuring kotlinx-benchmarks build against a custom dev version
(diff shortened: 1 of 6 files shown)


buildSrc/src/main/kotlin/KotlinCommunity.kt
```diff
@@ -37,4 +36,34 @@ fun addDevRepositoryIfEnabled(repositoryHandler: RepositoryHandler, project: Pro
...
+ */
+fun getOverriddenKotlinLanguageVersion(project: Project): String? {
+    val languageVersion = project.rootProject.properties["kotlin_language_version"] as? String
+    return if (!languageVersion.isNullOrBlank()) {
+        project.logger.info("""Configured Kotlin Language version: '$languageVersion' for project ${project.name}""")
+        languageVersion
+    } else {
+        null
+    }
+}
```
adam-enko, line 69:
```suggestion
fun getOverriddenKotlinApiVersion(project: Project): String? {
    val apiVersion = project.providers.gradleProperty("kotlin_api_version").orNull
    return if (!apiVersion.isNullOrBlank()) {
        project.logger.info("""Configured Kotlin API version: '$apiVersion' for project $${project.name}""")
        apiVersion
    } else {
        null
    }
}

/**
 * Should be used for running against non-released Kotlin compiler on a system test level.
 *
 * @return a Kotlin Language version parametrized from command line or gradle.properties, null otherwise
 */
fun getOverriddenKotlinLanguageVersion(project: Project): String? {
    val languageVersion = project.providers.gradleProperty("kotlin_language_version").orNull
    return if (!languageVersion.isNullOrBlank()) {
        project.logger.info("""Configured Kotlin Language version: '$languageVersion' for project ${project.name}""")
        languageVersion
    } else {
        null
    }
}
```

I recommend avoiding using the untyped Project properties and instead using the Provider API.

adam-enko (approved):

PR #268, opened 2024-10-17
Add a TC build config for building with Kotlin master

qurbonzoda:
Such a build configuration is useful for catching new changes that are not compatible with the latest Kotlin build.
For details see https://youtrack.jetbrains.com/articles/KT-A-297/Libraries-develop-and-Kotlin-master-checks

### DIFF
--- a/.teamcity/additionalConfiguration.kt
+++ b/.teamcity/additionalConfiguration.kt
@@ -106,8 +106,7 @@ fun Project.buildWithKotlinMaster(platform: Platform, versionBuild: BuildType) =
         artifacts {
             buildRule = lastSuccessful()
             cleanDestination = true
-            artifactRules = "+:maven.zip!**=>artifacts/kotlin\n" +
-                    "+:kotlin-native-prebuilt-linux-x86_64-%$kotlinVersionParameter%.tar.gz!**=>artifacts/native"
+            artifactRules = "+:maven.zip!**=>artifacts/kotlin"
         }
     }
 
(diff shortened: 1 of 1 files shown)


fzhinkin:
Is it possible to check that the new configuration will be created and will work as expected without merging this change into master?

.teamcity/additionalConfiguration.kt
```diff
@@ -85,4 +83,47 @@ fun Project.deployPlugin() = BuildType {
...
+        param(teamcitySuffixParameter, versionBuild.depParamRefs[teamcitySuffixParameter].ref)
+    }
+
+    val kotlinVersionParameter = "dep.Kotlin_KotlinPublic_BuildNumber.deployVersion"
+
+    dependsOn(AbsoluteId("Kotlin_KotlinPublic_Artifacts")) {
+        artifacts {
+            buildRule = lastSuccessful()
+            cleanDestination = true
+            artifactRules = "+:maven.zip!**=>artifacts/kotlin"
```
fzhinkin, line 109: Both AFU and Serialization are also copying `+:kotlin-native-prebuilt-linux-x86_64-%dep.Kotlin_KotlinPublic_BuildNumber.deployVersion%.tar.gz!** => artifacts/native`.

Does it make sense here as well (for sample projects)?

qurbonzoda: The unpacked `maven.zip` already contains the K/N distribution, and Gradle gets it from there. It is specified by `-Pkotlin_repo_url=<checkoutDir>/artifacts/kotlin`. I didn't find where and how `artifacts/native` is used, thus dropped it in the last commit.

qurbonzoda:
@woainikk has run kx-benchmark master with Kotlin master here: https://buildserver.labs.intellij.net/buildConfiguration/Kotlin_KotlinDev_BenchmarkCompilationCheckK2/588557805?logView=flowAware
It is successfully built. Tests were not executed though.

woainikk (approved):

fzhinkin (approved):

PR #285, opened 2025-01-07
Support Gradle Project Isolation solution.

deep4nsh:


plugin/main/src/kotlinx/benchmark/gradle/BenchmarkConfiguration.kt
```diff
@@ -70,7 +70,7 @@ constructor(
     extension: BenchmarksExtension,
     name: String
 ) : BenchmarkTarget(extension, name) {
-    var jmhVersion: String = (extension.project.findProperty("benchmarks_jmh_version") as? String) ?: "1.21"
+    var jmhVersion: String = providers.gradleProperty("benchmarks_jmh_version").orNull ?: "1.21"
```
fzhinkin, line 73: Does this code compiles?

fzhinkin:
Please add a regression test reproducing the original issue, if possible.
You can check tests in the `integration` module for examples.

PR #292, opened 2025-02-18
Fix assorted problems with K/JS projects

### LINKED ISSUE #185: Multiple issues when running on K/JS
Attached is a sample project to reproduce two issues. [benchmark-js-repro.zip](https://github.com/Kotlin/kotlinx-benchmark/files/14082248/benchmark-js-repro.zip)

1. Re-declaration of dependencies.  
    In the `benchmarks` module you're forced to re-declare all `core` dependencies and transitive dependencies.

2. Having files under `commonMain/resources` in `benchmarks` will block the entire build, with an error like:  
    ```
    Cause: e: Could not find "...\benchmark-js-repro\benchmarks\build\processedResources\js\main" in [...\benchmark-js-repro]`
    ```

fzhinkin:
Following issues revealed in #185 were addressed:
- dependency resolution problem was fixed so now there's no need to explicitly specify all transitive dependencies;
- having files in a resource directory no longer fail compilation;
- benchmarks compilation now inherit module kind from the main compilation.

Closes #185

### DIFF
--- a/runtime/jsMain/src/kotlinx/benchmark/NodeJsEngineSupport.kt
+++ b/runtime/jsMain/src/kotlinx/benchmark/NodeJsEngineSupport.kt
@@ -3,14 +3,19 @@ package kotlinx.benchmark
 import kotlin.time.DurationUnit
 import kotlin.time.toDuration
 
-internal external fun require(module: String): dynamic
+@JsModule("fs")
+@JsNonModule
+private external object NodeFileSystem {
+    fun writeFileSync(path: String, data: String)
+    fun readFileSync(path: String, options: String): String
+}
 
 internal object NodeJsEngineSupport : JsEngineSupport() {
     override fun writeFile(path: String, text: String) =
-        require("fs").writeFileSync(path, text)
+        NodeFileSystem.writeFileSync(path, text)
 
     override fun readFile(path: String): String =
-        require("fs").readFileSync(path, "utf8") as String
+        NodeFileSystem.readFileSync(path, "utf8")
 
     override fun arguments(): Array<out String> {
         val arguments = js("process.argv.slice(2).join(' ')") as String
--- a/integration/src/test/kotlin/kotlinx/benchmark/integration/KotlinJsTest.kt
+++ b/integration/src/test/kotlin/kotlinx/benchmark/integration/KotlinJsTest.kt
@@ -1,0 +1,26 @@
+package kotlinx.benchmark.integration
+
+import kotlin.test.Test
+
+class KotlinJsTest : GradleTest() {
+    @Test
+    fun useEsModules() {
+        project("es-modules", true).runAndSucceed("jsEsBenchmark") {
+            assertOutputContains("CommonBenchmark.benchmark")
+        }
+    }
+
+    @Test
+    fun useUmdModules() {
+        project("es-modules", true).runAndSucceed("jsUmdBenchmark") {
+            assertOutputContains("CommonBenchmark.benchmark")
+        }
+    }
+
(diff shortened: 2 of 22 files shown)


runtime/jsMain/src/kotlinx/benchmark/NodeJsEngineSupport.kt
```diff
@@ -3,14 +3,19 @@ package kotlinx.benchmark
 import kotlin.time.DurationUnit
 import kotlin.time.toDuration
 
-internal external fun require(module: String): dynamic
+@JsModule("fs")
```
lppedd, line 6: This makes sense since the docs say kx-benchmark only supports `nodejs()`.

However, I see the library also has a `D8EngineSupport`, so doesn't this change mean a require or import of `fs` will always be performed eagerly? Unlike it was done before - lazily?

fzhinkin: (Lucky me) Right now, D8 is not supported for K/JS: https://github.com/Kotlin/kotlinx-benchmark/blob/da3e2bbad26f5b531af8c9c39431ca1933ed525b/plugin/main/src/kotlinx/benchmark/gradle/JsEngineExecTasks.kt#L45
And for K/WasmJs, we use a different executor at runtime.

igoriakovlev: BTW, is it possible easily to support D8 for K/JS (for JB personal use)?

fzhinkin: Sure, we can try.

lppedd: @fzhinkin might have better info, but my understanding is you'd need [KT-20679](https://youtrack.jetbrains.com/issue/KT-20679).

The fact is:
1. annotating externals with `@JsModule` translates into eager resolution
2. custom lazy loading means you'd have to know when CommonJS is used to `require()`, or when ESM is used to use  dynamic `import()`

Apart from D8, in the future people might want to run benchmarks in a standard browser environment, but yeah, we seem to be limited now.

Might workaround limitation 2 by having a build-time constant inject, like

```kt
// Injected by a compiler plugin?
const val IMPORT: String = "require('...')" // or import('...')

// At use site
js(IMPORT).writeFileSync(path, text)
```

igoriakovlev: > Sure, we can try.

Cool, especially if we can do that with the same executor as `Wasm` do. We need it to compare `JS` and `WasmJS` performance. But it is not priority thing anyway.

fzhinkin: Filed https://github.com/Kotlin/kotlinx-benchmark/issues/293 to track the proposal.

fzhinkin: > but my understanding is you'd need [KT-20679](https://youtrack.jetbrains.com/issue/KT-20679).

We know the underlying runtime before generating the code, so we may try generating some part of the executor with appropriate import/require statements along with the benchmarks itself.

lppedd: Yeah that sounds much better compared to working around it with custom lazy loading.

JSMonk (approved):

PR #295, opened 2025-03-05
Update publish-plugin to published signed plugin artifacts

### LINKED ISSUE #115: Artifacts should be signed for dependency verification
In AndroidX we are needing to add exceptions for dependency verification for kotlinx-benchmark artifacts.

```
  These files failed verification:
    - GRADLE_USER_HOME/caches/modules-2/files-2.1/org.jetbrains.kotlinx.benchmark/org.jetbrains.kotlinx.benchmark.gradle.plugin/0.4.8/24d8d69982d768bcfa9c2f4c8aed7fdcec2ae5c3/org.jetbrains.kotlinx.benchmark.gradle.plugin-0.4.8.pom
    - GRADLE_USER_HOME/caches/modules-2/files-2.1/org.jetbrains.kotlinx/kotlinx-benchmark-plugin/0.4.8/452ca78e63a94ede9eb8b9dcfcc6c693708a1b22/kotlinx-benchmark-plugin-0.4.8.jar
    - GRADLE_USER_HOME/caches/modules-2/files-2.1/org.jetbrains.kotlinx/kotlinx-benchmark-plugin/0.4.8/dd0525949f4c23c9468860892d7ed4625a4834b4/kotlinx-benchmark-plugin-0.4.8.module
```

Repro:
1. `git clone git@github.com:androidx/androidx.git`
2. Manually remove `kotlinx.benchmark` entries in `gradle/verification-metadata.xml`
3. `cd collection && ./gradlew collection:collection-benchmark-kmp:assembleBenchmarks`

fzhinkin:
Updated infra and published plugins to unlock signed plugin artifacts publication.

Closes #115

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -4,10 +4,10 @@
 kotlin = "2.0.20"
 kotlin-for-gradle-plugin = "2.0.20" # Kotlin 2.1 removes support for the used language version / api version: KT-60521
 kotlinx-binaryCompatibilityValidator = "0.16.2"
-kotlinx-teamInfra = "0.4.0-dev-81"
+kotlinx-teamInfra = "0.4.0-dev-84"
 squareup-kotlinpoet = "1.3.0"
 jmh = "1.21"
-gradle-pluginPublish = "0.21.0"
+gradle-pluginPublish = "1.3.1"
 
 # Note: This version can be overridden by passing `-Pmin_supported_gradle_version=<version>`
 minSupportedGradle = "7.4"
--- a/plugin/build.gradle.kts
+++ b/plugin/build.gradle.kts
@@ -52,19 +52,17 @@ repositories {
     }
 }
 
-pluginBundle {
+gradlePlugin {
     website = "https://github.com/Kotlin/kotlinx-benchmark"
     vcsUrl = "https://github.com/Kotlin/kotlinx-benchmark.git"
-    tags = listOf("benchmarking", "multiplatform", "kotlin")
-}
 
-gradlePlugin {
     plugins {
         register("benchmarkPlugin") {
-            id = "org.jetbrains.kotlinx.benchmark"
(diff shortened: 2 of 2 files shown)


fzhinkin:
I tried to test it by changing plugin id and publishing it from my personal account. Obviously, verification on the portal side failed and the submission was rejected, but looking into Gradle's logs it seems like `.asc` files for plugin were uploaded.

@adam-enko, @Tapchicoma maybe you guys know a better testing approach?

adam-enko:
Looking at the PR, imho it'd be better to update kotlinx team infra plugin to avoid such workarounds. The last time I looked at the infra plugin it used a lot of non-idiomatic practices that could be refactored to avoid these issues. Is updating the infra plugin feasible?

fzhinkin:
> it'd be better to update kotlinx team infra plugin to avoid such workarounds.

I considered that option, but there's only a few projects still using the plugin and I though it would be easier to update one of them to exclude the duplicated artifact instead of updating all other projects to include it explicitly.

> the infra plugin it used a lot of non-idiomatic practices that could be refactored to avoid these issues

It makes sense to do all that at some point and then get rid of the workaround I'm adding here.

fzhinkin:
> Personally I test publishing by adding a local directory and manually verifying the files.

Apparently, signature files are always published to a local repo, no matter what version of publish plugin I'm using :/

adam-enko:
I just saw another user posted in the Gradle Slack with the same problem. Possibly there's an issue with the Plugin Portal?

image

https://gradle-community.slack.com/archives/CA745PZHN/p1741602255845399 ([archive](https://www.linen.dev/s/gradle-community/t/27066336/hello-peeps-i-got-a-request-of-https-github-com-gmazzo-gradl#b2854c21-f85f-44b6-bbe4-84252dc1af0c))

ilya-g (approved):

PR #309, opened 2025-05-14
Make WasmJS/D8 tasks configuration-cache friedly

fzhinkin:


plugin/main/src/kotlinx/benchmark/gradle/ConsoleAndFilesOutputStream.kt
```diff
@@ -65,4 +65,4 @@ internal class ConsoleAndFilesOutputStream : OutputStream() {
         buffer.close()
         currentStream.close()
     }
-}
```
igoriakovlev, line 68: useless change

igoriakovlev (approved):

plugin/main/src/kotlinx/benchmark/gradle/ConsoleAndFilesOutputStream.kt
fzhinkin: Dropped it, thanks for pointing out!

PR #318, opened 2025-07-18
[KUP] Disable opt-in Kotlin compiler warnings

FenstonSingel:
It was decided to deprecate and phase out the concept (see KT-77721).

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -107,12 +107,7 @@ allprojects {
             freeCompilerArgs.addAll("-Xreport-all-warnings", "-Xrender-internal-diagnostic-names")
 
             progressiveMode = true
-
-            if (getAllWarningsAsErrorsValue(project)) {
-                allWarningsAsErrors = true
-            } else {
-                freeCompilerArgs.addAll("-Wextra", "-Xuse-fir-experimental-checkers")
-            }
+            allWarningsAsErrors = getAllWarningsAsErrorsValue(project)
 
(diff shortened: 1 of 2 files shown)


FenstonSingel:
IDK what the problem with the TC build is, but it doesn't seem to be related to the changes given that [it's reproducible for the `master` branch](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxBenchmark_Build_with_Kotlin_Master_Linux/5340307).

woainikk (approved):

FenstonSingel:
> The kotlin-community/dev and master in benchmarks have differences, so the configs may also differ

AFAICT, [the only effective difference](https://github.com/Kotlin/kotlinx-benchmark/compare/master...kotlin-community/dev) is 89fde29dc5b9d6ea84de3bda3144f1a14bf6b1c7, and this difference _is_ accounted for in the public TC configuration by passing `-Pmin_supported_gradle_version=7.6.3` via the command line. So IDK, it still seems suspicious that our internal configuration is silent.

fzhinkin (approved):

fzhinkin:
> Native compilation fails for some reason (and I can't reproduce the failure outside CI)

It seems to be induced by Linux-only K/N builds we're using in a corresponding CI task. I disabled cross-compilation for that task and filed https://github.com/Kotlin/kotlinx-benchmark/issues/320 to enable it back later.

PR #321, opened 2025-08-04
Fix local fle paths in kotlin_repo_url on Windows

cypressious:


integration/src/main/kotlin/kotlinx/benchmark/integration/ProjectBuilder.kt
```diff
@@ -26,15 +26,15 @@ benchmark {
...
 }.orEmpty()
 
 private val plugin_repo_url = System.getProperty("plugin_repo_url")!!.let {
-    "maven { url '$it' }"
+    "maven { url '${it.replace("\\","\\\\")}' }"
 }
 
 private val runtime_repo_url = System.getProperty("runtime_repo_url")!!.let {
-    "maven { url '$it' }"
+    "maven { url '${it.replace("\\","\\\\")}' }"
```
fzhinkin, line 37: @cypressious could you please elaborate what problem you're trying to solve?

PR #327, opened 2025-10-10
Add examples of using Kotlin Notebook to analyze benchmark runs

cmelchior:
Two examples were added:
- "Simple analysis": Plot data from a single benchmark run
- "Compare analysis": Compare two runs of the same benchmark suite to find changes that are worth investigating further.

How these notebooks look can be seen here:

1. Simple: https://gist.github.com/cmelchior/1b6007752333d1ce76a1d02ed311a135
2. Compare: https://gist.github.com/cmelchior/76e87e5207a87c5c4968c31a8910a88a

Some questions for reviewers:
1. Are the notebooks easy to follow?
2. Are there too many comments? Too few?
3. Should we merge some of the intermediate steps? Currently, I try to make it more explicit what happens, but this also makes the text more verbose.
4. Do the math "checks out", i.e. is my way of using the data sensible or do you think something else would be more appropriate?


cmelchior:
Great, thanks 👍 

I updated the notebooks with your and @koperagen's comments.

The updated notebooks are here:
- https://github.com/Kotlin/kotlinx-benchmark/blob/78013f076439465ff6b1b6f711761c328953e57e/examples/compare-benchmark-runs.ipynb
- https://github.com/Kotlin/kotlinx-benchmark/blob/78013f076439465ff6b1b6f711761c328953e57e/examples/simple-benchmark-analysis.ipynb

fzhinkin (approved):
@cmelchior, thank you! I'll update the readme and start suggesting notebooks as part of plugin's output in a follow-up PR.

PR #330, opened 2025-10-30
Add information about analysis notebooks to README and print a link along with results

fzhinkin:
Also, updated notebooks to be more flexible with params.

### DIFF
--- a/runtime/commonMain/src/kotlinx/benchmark/BenchmarkProgress.kt
+++ b/runtime/commonMain/src/kotlinx/benchmark/BenchmarkProgress.kt
@@ -98,6 +98,7 @@ class ConsoleBenchmarkProgress : BenchmarkProgress() {
         println()
         println("$suite summary:")
         println(summary)
+        println("Analysing benchmark results is not always trivial,\nplease refer to https://github.com/Kotlin/kotlinx-benchmark/tree/master?tab=readme-ov-file#analyzing-results for details.\n")
     }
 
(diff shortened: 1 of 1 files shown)


fzhinkin:
@cmelchior could you please help with notebooks: I was going to add yet another notebook that'll compare benchmarks from the same run with each other and stumbled across a problem I can't solve quickly.

The JSON file with run results may not contain the `param` attribute for benchmarks, and in this case I don't understand how to inject some default value (like `null`), so that the rest of a notebook will continue to work without any changes.

For instance, consider plugging such a result file into a notebook: https://gist.githubusercontent.com/fzhinkin/afb9423170056af2696877f837f7950b/raw/31f6bb2e45ba5888f0b78ad7ebb9e62d5be563b0/main.json

Maybe you have some good ideas?

fzhinkin:
@cmelchior, the second option sounds good to me. After all, later all that could be moved to a separate artifacts that notebooks will use as a dependency, I guess.

> For the example file you had there, I copied a notebook and did minimal adjustments to remove it, trying to parse the params:
https://gist.github.com/cmelchior/84e4f13a20e7204202d63b2ee4f151ec

It seems like it does not work :(

fzhinkin:
Slightly trimmed the `Benchmark` DTO, so it now works with both JMH format and a simplified version of this format used for other targets.

Also, added an example project comparing some "baseline" and "optimized" implementations of the _same_ algorithm, and added a notebook that analyses its results.

ilya-g (approved):

Issue #358, opened 2026-03-26
Support multithreaded benchmarks


fzhinkin:
Currently, kotlinx-benchmark does no provide neither API, nor support a runtime feature to execute a benchmark function in parallel using the specific number of threads.

On JVM, one can either use JMH annotations, or pass a corresponding argument to the generated benchmark jar to run a benchmark in parallel.
On JS and Wasm, there are no threads.
However, on Native, there are threads and there's no way to utilize more than one to run a benchmark.

In certain scenarios (like https://github.com/Kotlin/kotlinx-atomicfu/issues/584), it's a benchmark has to be executed by multiple threads in parallel and the only workaround is to write infrastructure for that from scratch.

The niche for such benchmarks seems to be pretty small, yet crucial - we need it to ensure performance characteristics of different concurrent APIs provided by either kotlinx-libraries, or Kotlin stdlib.

The proposal is to add support of multithreaded benchmarks to kotlinx-benchmark.

There should be a common API to specify the number of threads. However, if the underlying platform is single-threaded (like JS), the plugin has to warn about it when running a benchmark.

### Multithreaded benchmarks feature list
- [ ] Annotations to specify the number of threads (like `@Threads` in JMH)
  - [ ] Class-level annotations
  - [ ] Function-level annotations
- [ ] DSL to specify the number of threads for existing benchmarks
- [ ] Symmetrical multithreaded benchmarks support
- [ ] Asymmetrical multithreaded benchmarks support (a JMH-inspired feature)
- [ ] Thread-local benchmark state and thread-local setup and teardown functions

Issue #179, opened 2024-01-14
Native benchmark: iteration counter overflow


fzhinkin:
During the warmup phase, the native benchmarks executor estimates the repetitions of benchmarked function calls within each measurement iteration. That number is stored in an integer variable. Given a fast enough benchmark and long enough `iterationTime`, the repetitions counter (called `iterations` in the code) may overflow, leading to negative results in a report.

```
@State(Scope.Benchmark)
@BenchmarkMode(Mode.AverageTime)
open class TestBenchmark {
    private var x = 123L

    fun nextValue(): Long {
        val r = x
        x = -1 * (x * 6364136223846793005 + 1442695040888963407)
        return r
    }
    
    @Benchmark
    fun signBenchmark(): Int = nextValue().sign    
}
```

```
    configurations {
        named("main") {
            warmups = 2
            iterations = 3
            iterationTime = 40 * 4
            iterationTimeUnit = "sec"
            outputTimeUnit = "ns"
        }
    }
```

Results:
```
Warm-up #0: -79-0.538163 ns/op
Warm-up #1: -81-0.545774 ns/op
Iteration #0: 0-0.000000 ns/op
Iteration #1: 0 ns/op
Iteration #2: 0-0.000000 ns/op
  Success:   ~ 0-0.000000 ns/op ±-98-0.00%
```

Issue #177, opened 2024-01-10
Native benchmark: executor incorrectly estimates the number of repetitions within each measurement iteration


fzhinkin:
During the warmup phase, the native benchmarks executor estimates the number of repetitions for each measurement iteration (i.e. the number of times the inner loop should invoke a benchmarked function).
Unfortunately, for short-running benchmarks, the machinery required for the warmup adds an overhead that skews the average time per benchmark invocation. That causes the estimated repetitions number to be too low, and as a result, the executor spends significantly less time in each measurement iteration than a user configured.

Consider the following example (https://github.com/fzhinkin/kt-64361-benchmarks/blob/main/kmp-benchmarks/src/commonMain/kotlin/SignumBenchmarks.kt):
```
… org.example.LongSignumBenchmark.signBitExtractingSignum
Warm-up #0: 24.6405 ns/op
Warm-up #1: 24.7149 ns/op
Warm-up #2: 24.7893 ns/op
Warm-up #3: 24.6666 ns/op
Warm-up #4: 24.6982 ns/op
Iteration #0: 3.97785 ns/op
Iteration #1: 3.99065 ns/op
Iteration #2: 3.95236 ns/op
...
```

As you can see, the average time reported during warmup is 6 times lower compared to what was measured later.
Not sure how to represent it in a textual form, but if you run the benchmark locally and watch the log messages being printed, you'll notice that each measurement iteration takes significantly less time compared to the warmup.
In my case, the duration of a single iteration was set to 1 second, and during warmup, it felt like it took about that time for each iteration. But when it comes to measurement, each iteration ends almost instantly.

The problem here is that the warmup phase makes a call to get a current timestamp after every single benchmark method call. For long-running benchmarks, it's hard to notice a difference, but for short-running benchmarks, it adds a significant overhead.

As it may not be possible to create a separate thread performing the warmup so it could be interrupted by the timer on all platforms (or we can simply create a pthread?), maybe we can adjust the procedure to check the elapsed time every N benchmarks calls?

fzhinkin:
The same issue is also present in WasmJs executor

qurbonzoda:
Related issue: https://github.com/Kotlin/kotlinx-benchmark/issues/19

PR #359, opened 2026-03-26
Basic multithreaded benchmarks support

### LINKED ISSUE #19: Weird result for warmup vs actual iterations for native benchmarks
This is the benchmark class:

```
@State(Scope.Benchmark)
@OutputTimeUnit(BenchmarkTimeUnit.MILLISECONDS)
@Measurement(time = 1, timeUnit = BenchmarkTimeUnit.SECONDS)
@BenchmarkMode(Mode.Throughput)
class SimpleOps {
  private var a = 0
  private var b = 0

  @Setup
  fun setup() {
    a = 10
    b = 20
  }

  @Benchmark
  fun add(): Int {
    return a + b
  }

  @Benchmark
  fun sub(): Int {
    return a - b
  }

  @Benchmark
  fun mul(): Int {
    return a * b
  }

  @Benchmark
  fun div(): Int {
    return a / b
  }
}
```

The results:

```
… benchmark.SimpleOps.add
Warm-up #0: 6,829.74 ops/ms
Warm-up #1: 7,094.42 ops/ms
Iteration #0: 89,446.7 ops/ms
Iteration #1: 92,966.6 ops/ms
Iteration #2: 83,1251.0 ops/ms
  Success:   ~ 88,513.1 ops/ms ±6.4%

… benchmark.SimpleOps.div
Warm-up #0: 6,988.84 ops/ms
Warm-up #1: 6,935.19 ops/ms
Iteration #0: 69,1761.0 ops/ms
Iteration #1: 69,718.8 ops/ms
Iteration #2: 82,222.2 ops/ms
  Success:   ~ 73,706.0 ops/ms ±11%

… benchmark.SimpleOps.mul
Warm-up #0: 5,537.26 ops/ms
Warm-up #1: 5,530.41 ops/ms
Iteration #0: 21,964.3 ops/ms
Iteration #1: 22,796.2 ops/ms
Iteration #2: 21,789.4 ops/ms
  Success:   ~ 22,183.3 ops/ms ±2.7%

… benchmark.SimpleOps.sub
Warm-up #0: 7,040.06 ops/ms
Warm-up #1: 7,041.18 ops/ms
Iteration #0: 91,877.2 ops/ms
Iteration #1: 78,932.6 ops/ms
Iteration #2: 90,539.2 ops/ms
  Success:   ~ 87,116.3 ops/ms ±9.2%
```

As you can see the actual iteration values is 10x of the warmup. It's probably due to this: https://github.com/Kotlin/kotlinx-benchmark/blob/master/runtime/nativeMain/src/kotlinx/benchmark/native/NativeExecutor.kt#L92 Before dividing should we change time to be the unit specified by the benchmark?

fzhinkin:
Added class-level `@Threads` annotation: it accepts either a positive integer number which will be the number of threads running a benchmark, or a special constant `THREADS_CPU_COUNT = -1`, that'll hint the executor to use the number of CPU cores as the number of threads.

The number of threads could be also set through Gradle DSL, using `BenchmarkConfiguration`'s `threads` property.

The number of threads used for running a benchmark is included into generated reports.

On JVM, the `@Threads` annotation is mapped directly to JMH's one. On Wasm and Js it is effectively no-op.

On Native, executor spawns `@Threads.value` `Worker`s and run every benchmark function within each of them. Iteration start is synchronized using a barrier, after all threads arrive, they start running the benchmark function in parallel. Benchmark class' instance is shared across all threads. A thread-local scope for a `@State` is out of the scope of this PR.

For simplicity, a single-threaded benchmarks are also executed in workers.

Along the way, I removed old logic responsible for estimating the number of benchmark function calls corresponding to specified iteration duration. Such a procedure was known for incorrect estimates (https://github.com/Kotlin/kotlinx-benchmark/issues/177, https://github.com/Kotlin/kotlinx-benchmark/issues/19). Instead, benchmarks are now executed in a loop until a volatile flag is updated, and a main thread is responsible for updated it once a timer expired.

Benchmark results are then aggregated in the JMH style:
- throughput is calculated as the total number of benchmark function's calls across all threads per specified unit of time
- average time is first calculated for each thread individually, and then averaged across them all.

Fixes: #19, #179
Partially addresses: #358


runtime/commonMain/src/kotlinx/benchmark/CommonBenchmarkAnnotations.kt
```diff
@@ -98,4 +98,18 @@ expect annotation class Measurement(
...
+ * [value] has to be either strictly positive (> 0), or it could be a special constant [THREADS_CPU_COUNT].
+ */
+@Target(AnnotationTarget.CLASS)
+expect annotation class Threads(val value: Int)
+
+/**
+ * A special value for [Threads.value] forcing to run a benchmark using `N`-threads,
+ * where `N` is the total number of CPUs available in runtime.
+ */
+const val THREADS_CPU_COUNT: Int = -1
```
whyoleg, line 115: Just thinking out loud: it's a bit sad that we can't have static (companion) property `MAX` on `Threads` as it exists in JMH.
Related to https://youtrack.jetbrains.com/issue/KT-29882/Support-actualizing-expect-class-containing-companion-with-Java-class-having-matching-statics.
Curious if it’ll be possible to support this (or if it's planned) when companion blocks will be supported. If yes, in future we could have: 
```kotlin
expect annotation class Threads(val value: Int) {
    companion {
        val MAX: Int
    }
}
```
In this case, no top-level `THREADS_CPU_COUNT` property will be needed at all...

runtime/jsMain/src/kotlinx/benchmark/js/JsBenchmarkExecutor.kt
```diff
@@ -90,6 +92,9 @@ class JsBenchmarkExecutor(name: String, @Suppress("UNUSED_PARAMETER") dummy_args
 
                 jsBenchmark.on("start") { _ ->
                     reporter.startBenchmark(executionName, id)
+                    if (warnAboutThreadsCount) {
+                        reporter.output(executionName, id, SINGLE_THREADED_WARNING)
```
whyoleg: Q: why it's implemented as a runtime warning, and not inside `SuiteSourceGenerator`, or more generally at build time in Gradle plugin?
In the same way, if we could enforce `1` value for js/wasm-js there.

Also, it might make sense to be able to suppress this warning in some way, as if the project has multiplatform benchmarks, it's kind-a expected? WDYT?

runtime/nativeMain/src/kotlinx/benchmark/native/NativeExecutor.kt
```diff
@@ -71,24 +87,31 @@ class NativeExecutor(
...
+
+        WorkersPool(benchmarkRun.config.threads).use { workersPool ->
+            try {
+                warmup(
+                    benchmark.suite.name, benchmarkRun.config, instance,
+                    benchmark,
+                    workersPool,
+                    iteration.toInt()
+                )
+                resultsFile.writeFile("1") // Iterations number for backward compatibility
```
whyoleg: just curious: compatibility with what?

whyoleg (commented):
Just some random comments. The implementation feels like a black box to me, sorry :(

runtime/nativeMain/src/kotlinx/benchmark/native/NativeExecutor.kt
fzhinkin: The Gradle plugin - it reads this info from the result file and then pass to the measurements runner. It all matters only if runtime and plugin's versions do not match. And if someone uses `perIteration` forking mode, for some reason.

fzhinkin: But I broke the per-iteration for mode anyway :|

runtime/jsMain/src/kotlinx/benchmark/js/JsBenchmarkExecutor.kt
fzhinkin: > Q: why it's implemented as a runtime warning, and not inside `SuiteSourceGenerator`, or more generally at build time in Gradle plugin?

I moved the check to runtime to handle thread number values coming from both annotation and the config in the build script, but it all should be done on the plugin level instead, you're right.

> Also, it might make sense to be able to suppress this warning in some way, as if the project has multiplatform benchmarks, it's kind-a expected? WDYT?

That makes perfect sense, thanks!

fzhinkin: Well, it managed to add logging to Gradle, but it all so awkward, that I'm considering removing the warning altogether.

whyoleg: Well, I agree!

Another suggestion: what if we always print the specified number of threads, as well as the number of threads currently available for the benchmark, when `threads != 1`? Not only for JS, but in all cases.

E.g., JMH by default outputs `Threads`, and it looks fine:
```
# Warmup: 5 iterations, 10 s each
# Measurement: 5 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Throughput, ops/time
...
```

So, in our case, if the user specified `threads != 1` via annotation or configuration property, we will print something like this before every test run, before warm-up/iteration outputs, after :
```
Running 'main' benchmarks for 'jvm'
jvm: dev.whyoleg.cryptography.serialization.pem.benchmarks.PemBenchmark.decodeFromByteString | testCaseString=privateKey
Threads: 5 (max runtime threads = 1)

Warm-up 1: 1979.368 ops/ms
```

E.g., after some more thinking about it, the warning isn't really specific to JS, and maybe it makes sense to always show the number of current machine threads to give users a hint that running benchmarks for 10 threads on a 4-threaded machine isn't a good idea?
WDYT?

If not, we can still just print `Threads: 5`, but on JS add `(single-threaded runtime)` or similar. :)

runtime/nativeMain/src/kotlinx/benchmark/native/NativeExecutor.kt
```diff
@@ -218,115 +256,159 @@ internal class NativeExecutor(
...
-    private fun <T> warmup(
-        name: String,
-        config: BenchmarkConfiguration,
+    private inline fun singleIterationLoop(
+        synchronizer: MeasurementSynchronizer,
+        nativeGCAfterIteration: Boolean,
+        body: () -> Unit,
+    ): IterationResult {
+        if (nativeGCAfterIteration)
+            GC.collect()
```
whyoleg: Currently, we do call `GC.collect` inside each worker - shouldn't we move this check to before we submit the tasks to the worker pool?
Otherwise, it seems we might call GC before the iteration in another worker has already started.
Not sure if it will make a big difference, though.

A similar situation could happen with the second GC.collect`, which could also affect the last cycle of other workers.

runtime/appleMain/src/synchronizationApple.kt
```diff
@@ -0,0 +1,75 @@
...
+    private val mutex: pthread_mutex_t = arena.alloc()
+    private val cond: pthread_cond_t = arena.alloc()
+
+    private var closed = false
+    private var allArrived = false
+
+    init {
+        pthread_cond_init(cond.ptr, null).let {
+            if (it != 0) {
+                closed = true
```
whyoleg: Doesn't makes sense to set flag in `init` block, as it will never be observed

runtime/appleMain/src/synchronizationApple.kt
```diff
@@ -0,0 +1,75 @@
...
+            }
+        }
+    }
+
+    actual fun wait(): Unit {
+        check(!closed) { "Barrier was closed already" }
+        check(!allArrived) { "Barrier was already reached by the required number of parties." }
+
+        pthread_mutex_lock(mutex.ptr).let {
+            check(it == 0) { "Lock acquisition failed: $it" }
```
whyoleg: curious: why it uses `check`, while everywhere else you just check for `it != 0` - both do the same thing right? :)
but reading two different usages forces me to think that there is a difference.

Also, while I was working in CK on error handling, I found it's convenient to have a function like `checkError(Int, (Int) -> String): Int` to reduce boilerplate on error handling of native calls - might be  nice to have something like this.
At least for me, the pattern with `call.let {...}` feels too verbose.

runtime/nativeMain/src/kotlinx/benchmark/native/Nanosleep.kt
```diff
@@ -0,0 +1,51 @@
...
+import kotlinx.cinterop.ptr
+import platform.posix.nanosleep
+import platform.posix.timespec
+import kotlin.ranges.contains
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
+
+@OptIn(UnsafeNumber::class, ExperimentalForeignApi::class)
+internal class Nanosleep(duration: Duration) : AutoCloseable {
+    val arena = Arena()
```
whyoleg: all properties could be private

runtime/nativeMain/src/kotlinx/benchmark/native/Nanosleep.kt
```diff
@@ -0,0 +1,51 @@
...
+    val timespecInitial: timespec = arena.alloc()
+    val timespecRemaining: timespec = arena.alloc()
+
+    init {
+        require(duration.isPositive())
```
whyoleg: BTW: the `arena` stays not closed if `init` block fails - shouldn't be an issue in real code, but, it's a memory leak.

runtime/nativeMain/src/kotlinx/benchmark/native/NativeExecutor.kt
```diff
@@ -218,115 +256,159 @@ internal class NativeExecutor(
...
+                // Synchronized workers
+                barrier.wait()
+
+                sleepWrapper.sleep()
+
+                // We're done
+                synchronizer.shouldStop = true
             }
+
+            return AggregateIterationResult(futures.map { it.result }.toTypedArray())
```
whyoleg: curious: is there a specific reason to convert it to an array, and not store as a list?

runtime/nativeMain/src/kotlinx/benchmark/native/NativeExecutor.kt
```diff
@@ -134,27 +155,39 @@ internal class NativeExecutor(
...
-                val iterationNumber = currentIteration ?: iteration
-                reporter.output(
-                    executionName,
-                    id,
-                    "Iteration #$iterationNumber: $text"
-                )
-                nanosecondsPerOperation.nanosToSample(benchmarkRun.config.mode, benchmarkRun.config.outputTimeUnit)
+        val iterations =
+            if (benchmarkRun.config.nativeFork == NativeFork.PerIteration) 1 else benchmarkRun.config.iterations
+        val samples = WorkersPool(benchmarkRun.config.threads).use { workersPool ->
```
whyoleg: It would be nice to compare single-threaded results before and after the change.
E.g., does the results looks similar (so that the logic is correct), and if the fact, that we now run everything via worker affects it in anyway.

whyoleg (changes requested):
One more pack of comments regarding the implementation, now after a bit more "in-depth" review, but I'm still processing the whole implementation.

Do you think we could split the PR (make it stacked?) for easier, faster review?

runtime/nativeMain/src/kotlinx/benchmark/native/Nanosleep.kt
```diff
@@ -0,0 +1,51 @@
...
+import kotlinx.cinterop.convert
+import kotlinx.cinterop.ptr
+import platform.posix.nanosleep
+import platform.posix.timespec
+import kotlin.ranges.contains
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
+
+@OptIn(UnsafeNumber::class, ExperimentalForeignApi::class)
+internal class Nanosleep(duration: Duration) : AutoCloseable {
```
whyoleg: BTW: the contract of the class usage is a bit strange, considering the fact that the time starts "ticking" at the moment of class creation. I would probably expect this to be just a function, like `waitUntil(Duration, () -> Unit)`, rather than a class, so it's harder to misuse. Though the fact that it's hard to find a good name for it, as well as the fact that it's an implementation detail for a very specific usage, might not be worth it.

But while thinking about it, I've also taken a look at the usage and, well, thought that it could be simplified by encapsulating more into `MeasurementSynchronizer` than just the flag: https://github.com/Kotlin/kotlinx-benchmark/commit/2ee7397f22cd61e6d4b77ea196d6795160fe9aa7.

Feel free to discard or incorporate the idea if you see it fitting. :)

PR #366, opened 2026-04-10
Fix TeamCity build icon

cmelchior:
It looks like the default branch on TeamCity was set to `test-new-publication-scheme`, which broke the TeamCity build icon in the README.

I was a bit afraid to change the TeamCity configuration, so this commit forces the URL to fetch the status from `master`. 

I guess there is a chance we move away from `master` as the main branch, but it feels less likely than switching the default branch on TC.


fzhinkin (approved):
Thanks for fixing it!

PR #368, opened 2026-04-19
Add support for Android as a benchmark target

### LINKED ISSUE #116: Can it support android and ios targets?
We are developing a kmm project and want to compare the performance of android and androidNative. Does this benchmark test support the Android platform, or is there any other way?

### LINKED ISSUE #206: Unclear how to use this lib on android projects (for kotlin code)
I tried to follow the Kotlin/JVM setup https://github.com/Kotlin/kotlinx-benchmark/blob/master/docs/kotlin-jvm-project-setup.md but unfortunately I could not get it working. 

I have added the plugin & benchmark runtime dependency and added this to the `build.gradle`:

```
benchmark {
    targets {
        register("main") // also tried app.main
    }
}
```

I can import the annotations, but they are giving errors. 

Screenshot 2024-05-03 at 08 54 00

Running `./gradlew benchmark` returns `> Configure project :app     Warning: Cannot find a benchmark sourceSet 'main', ignoring.`

cmelchior:
Closes #206
Closes #116 

This PR replaces https://github.com/Kotlin/kotlinx-benchmark/pull/250, which never made it across the finish line and has now bit-rotted quite a bit. It updates all relevant API's and addresses all the issues raised in the original PR.

The base functionality is done, but there are still some open questions. These are described in [this doc](https://github.com/Kotlin/kotlinx-benchmark/blob/cm/android-support/android-support-questions.md). 

**Main changes compared to #250:**
- [x] Android Benchmarks can now be added in KMP projects using the `com.android.kotlin.multiplatform.library` plugin instead of using `com.android.library` or `com.android.application`.
- [x] Gradle tasks have been refactored to be Configuration Cache safe and now have correct Input/Output for better build performance.
- [x] The template project will now inherit as many options as possible from the parent project, and `AndroidBenchmarkTarget` has an API to configure it further.
- [x] The template project now correctly includes transitive dependencies from the main project.
- [x] The annotation processor and code generation were ported with minimal changes, fixing only some minor bugs (Thanks to @kx412764776 for the original code)
- [x] Kotlin Poet was updated to 2.3.0 to fix code generation wrapping lines in the wrong places.
- [x] Updated to Jetpack Microbenchmark version `1.5.0`, including using the new `measureRepeated` API.
- [x] Documentation has been added to the main README.
- [x] An Android-specific guide (similar to JVM) was added.
- [x] The `kotlin-multiplatform` example was extended to support Android as well.

**Outstanding issues:**
This is a list of known issues, but it should be clarified if all (or some) need to be fixed before this PR can be merged.

- [x] Figure out how to handle the JDK 11 issue.
- [ ] Determine how to handle testing on Team City now that the Android SDK is required. 
- [x] The annotation processor does not support passing in Blackhole (test is ignored):
        ```
         @Benchmark
          fun test(bh: Blackhole) { ... }
        ```
- [x] Parameterized tests produce invalid code in some cases (Can be reproduced on `kotlinx.collection.immutable` tests).
- [ ] A test suite for the Annotation Processor to ensure it works as expected.
- [x] Integration tests have to be expanded to test different versions of AGP.
- [x] AGP 9 no longer requires the `kotlin-android` plugin. Figure out how to handle user versions crossing this boundary.
- [ ] `@Warmup/@Measurement` is not supported.
- [ ] `@OutputTimeUnit` is not supported.
- [ ] Convert JSON output from Jetpack Microbenchmark to a JMH-compatible format.
- [x] Allow Gradle download URL to be configurable.
- [x] Make `dryRun` an explicit API in `AndroidBenchmarkTarget` instead of using `instrumentationRunnerArguments` options.
- [x] Expose `lockClocks`/`unlockClocks` through `kotlinx-benchmark` Gradle tasks.
- [x] Support for chunking benchmarks. `kotlinx.collections.immutable` creates 4000+ tests which crash gRPC. 
        Should be doable by running test pr class and then combine afterwards using
         ```
         ./gradlew connectedCheck -Pandroid.testInstrumentationRunnerArguments.class=...
         ```
- [ ] Respect properties set in `BenchmarkConfiguration` (or as many as possible)
    - [ ] `iterations` Jetpack does not support this natively 
    - [ ] `warmups` Jetpack does not support this natively
    - [ ] `iterationTime` Jetpack does not support this natively
    - [ ] `iterationTimeUnit` Jetpack does not support this natively
    - [ ] `mode` Jetpack does not support this natively
    - [ ] `outputTimeUnit` Jetpack does not support this natively, it always reports in ns.
    - [x] `reportFormat`
    - [x] `includes`
    - [x] `excludes`
    - [x] `params` Not supported right now
    - [ ] `advanced` Not supported right now
         
**Testing**
The current state of this PR was tested by:
- Running `kotlin-multiplatform` example.
- Running all tests in the `integration` module. 
- Added Android Benchmark support to `kotlinx.collections.immutable` and tested that these can run.


cmelchior:
Regarding tests. I updated the Team City setup to only run on nodes with the Android SDK, but it looks like some security settings prevent it from being used. So until it can be tested, that is just a best-guess implementation for what it could look like.

PR #373, opened 2026-05-07
Add compilation mode into benchmark report

### LINKED ISSUE #372: A report file to have a description or a custom user field
We found that it is not very suitable to have report files without any description inside, like is this report for development or production mode. With which wasm engine it was produced, what configuration have been used. So it is hard to discover how a report is produced. This is especially a problem when produced reports being consumed by performance dashboards.
We suggest either to have a custom field in report file or have such info in a static manner. Having possibility to name a report file by user could be also heplfull.

igoriakovlev:


plugin/main/src/kotlinx/benchmark/gradle/JsEngineExecTasks.kt
```diff
@@ -60,6 +62,7 @@ private fun MutableList<String>.addJsArguments() {
 private fun createNodeJsExec(
     config: BenchmarkConfiguration,
     target: BenchmarkTarget,
+    mode: KotlinJsBinaryMode,
     compilation: KotlinJsIrCompilation,
```
fzhinkin: Perhaps, we can replace it with binary and extract both mode and compilation in this function?

runtime/commonMain/src/kotlinx/benchmark/BenchmarkReportFormatter.kt
```diff
@@ -45,6 +49,8 @@ internal object TextBenchmarkReportFormatter : BenchmarkReportFormatter() {
                 appendPaddedBefore("($it)", paramLengths[it]!!)
             }
             appendPaddedBefore("Mode", modeLength)
+            compilationModeAndLength?.let { appendPaddedBefore("CompilationMode", it.second) }
+            appendPaddedBefore("Mode", modeLength)
```
fzhinkin: The second "Mode" is clearly a typo here

runtime/commonMain/src/kotlinx/benchmark/BenchmarkReportFormatter.kt
```diff
@@ -45,6 +49,8 @@ internal object TextBenchmarkReportFormatter : BenchmarkReportFormatter() {
                 appendPaddedBefore("($it)", paramLengths[it]!!)
             }
             appendPaddedBefore("Mode", modeLength)
+            compilationModeAndLength?.let { appendPaddedBefore("CompilationMode", it.second) }
```
fzhinkin: Do you really need the compilation mode anywhere but the json report file?
TBH, I thought it was all about adding extra info into the JSON report only.

igoriakovlev: I dont know. For us it is okey to have it only in json. Say WDYT!

runtime/commonMain/src/kotlinx/benchmark/BenchmarkReportFormatter.kt
igoriakovlev: Yeah. fixed

plugin/main/src/kotlinx/benchmark/gradle/JsEngineExecTasks.kt
igoriakovlev: done

runtime/commonMain/src/kotlinx/benchmark/BenchmarkReportFormatter.kt
fzhinkin: Yeah, let's keep it in JSON only. Looking at all other stuff we put into reports, everything goes into JSON and other formats report only bare minimum, so let's keep it that way.

igoriakovlev: Sure. I have removed this from the other formatters.

fzhinkin (approved):

PR #375, opened 2026-05-08
Get rid of redundant js functions, use Js Modules instead

ilgonmic:


fzhinkin:
@igoriakovlev, could you please check if your custom engines setup work correctly with this change?

igoriakovlev:
> @igoriakovlev, could you please check if your custom engines setup work correctly with this change?

@fzhinkin Thank for mentionining!
I do not have any idea what is going on there...looks scary :)
Nevertheless, I have tested this and seems everything works well for us.

fzhinkin (approved):

PR #378, opened 2026-05-21
Fix run suite which does not exists

igoriakovlev:
If someone wants to run a suite which exists on spawn runner but does not exists in runned suites set it should not fail but just be skipped

### DIFF
--- a/runtime/jsMain/src/kotlinx/benchmark/js/JsBuiltInExecutor.kt
+++ b/runtime/jsMain/src/kotlinx/benchmark/js/JsBuiltInExecutor.kt
@@ -17,7 +17,7 @@ fun runBenchmarksBuiltIn(name: String, args: Array<out String>, declareAndExecut
             SingleBenchmarkExecutor(
                 executionName = name,
                 runnerConfiguration = RunnerConfiguration(configPath.readFile()),
-                suiteIndex = arguments[1].toInt(),
+                suiteId = arguments[1],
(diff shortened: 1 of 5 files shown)


fzhinkin:
Is it possible to add a regression test for it? Otherwise, the change looks good.

igoriakovlev:
> Is it possible to add a regression test for it? Otherwise, the change looks good.

Well, we could to make that but the whole thing is about very odd thing - this issue fires only when we ask to enumerate one target bench set with the another target bench set. (what we do in Wasm). I do not think it is what someone supposed to do :)

fzhinkin:
Could you please share the reproducer, at least? So we could refer to it later. Right now, while the change looks safe, it is somewhat obscure to me.

igoriakovlev:
[suite_id_invalid.zip](https://github.com/user-attachments/files/28517809/suite_id_invalid.zip)
This is reproducer. To see the problem you need to run:
`./gradlew :compileWasmWasiBenchmarkProductionExecutableKotlinWasmWasiOptimize`
`./gradlew :wasmJsBenchmarksBenchmark`
and before this MR you will see that there is errors in the console and there is no `c.c_Benchmarks.c_Benchmark` in the output (althought it exists in common code). After this commit all wasi benchmarks for are executed.

fzhinkin (approved):

fzhinkin (approved):

PR #380, opened 2026-06-11
Do not use JSPI in Wasm runtime

igoriakovlev:
In kotlin.wasm we are going to redesign js part of wasm app. So wasmExports will not be available to use from kotlin. That means that JSPI not cannot be configured from Kotlin.

Here, I remake the way how we wait for a process in a wasm perBenchmark mode to avoid JSPI.
I made a special nodejs process which is intermediate between benchmark engine and benchmark runner, which waits for other process and splits the result from console output into a special output pipe.

### DIFF
--- a/runtime/jsMain/src/kotlinx/benchmark/js/JsBuiltInExecutor.kt
+++ b/runtime/jsMain/src/kotlinx/benchmark/js/JsBuiltInExecutor.kt
@@ -6,22 +6,26 @@ import kotlinx.benchmark.internal.KotlinxBenchmarkRuntimeInternalApi
 @KotlinxBenchmarkRuntimeInternalApi
 fun runBenchmarksBuiltIn(name: String, args: Array<out String>, declareAndExecuteSuites: (SuiteExecutorBase) -> Unit) {
     val arguments = engineSupport.arguments()
-    val configPath = arguments[0]
 
-    val executor = when (arguments.size) {
-        2 -> {
-            check(arguments[1] == "startAll")
+    val executor = when {
+        arguments.any { it == "startAll" } -> {
+            val newArguments = arguments.filterNot { it == "startAll" }.toTypedArray()
+            val configPath = newArguments[0]
             JsBuiltInExecutor(name, configPath)
         }
-        3 -> {
+        arguments.any { it == "runSingle" } -> {
+            val newArguments = arguments.filterNot { it == "runSingle" }.toTypedArray()
+            val configPath = newArguments[0]
+            val config = RunnerConfiguration(configPath.readFile())
             SingleBenchmarkExecutor(
                 executionName = name,
-                runnerConfiguration = RunnerConfiguration(configPath.readFile()),
+                runnerConfiguration = config,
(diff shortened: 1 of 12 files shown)


runtime/jsMain/src/kotlinx/benchmark/js/JsBuiltInExecutor.kt
```diff
@@ -6,22 +6,26 @@ import kotlinx.benchmark.internal.KotlinxBenchmarkRuntimeInternalApi
...
-        3 -> {
+        arguments.any { it == "runSingle" } -> {
+            val newArguments = arguments.filterNot { it == "runSingle" }.toTypedArray()
+            val configPath = newArguments[0]
+            val config = RunnerConfiguration(configPath.readFile())
             SingleBenchmarkExecutor(
                 executionName = name,
-                runnerConfiguration = RunnerConfiguration(configPath.readFile()),
+                runnerConfiguration = config,
                 suiteId = arguments[1],
```
fzhinkin: Should it `newArguments` instead?

runtime/jsMain/src/kotlinx/benchmark/js/JsBuiltInExecutor.kt
```diff
@@ -6,22 +6,26 @@ import kotlinx.benchmark.internal.KotlinxBenchmarkRuntimeInternalApi
...
     val arguments = engineSupport.arguments()
-    val configPath = arguments[0]
 
-    val executor = when (arguments.size) {
-        2 -> {
-            check(arguments[1] == "startAll")
+    val executor = when {
+        arguments.any { it == "startAll" } -> {
+            val newArguments = arguments.filterNot { it == "startAll" }.toTypedArray()
+            val configPath = newArguments[0]
```
fzhinkin: `val configPath = arguments.first { it != "startAll" }` ?

runtime/wasmWasiMain/src/kotlinx/benchmark/wasm/WasmWasiExecutorFactory.kt
```diff
@@ -7,18 +7,16 @@ import kotlinx.benchmark.wasm.wasi.wasiGetArguments
 @KotlinxBenchmarkRuntimeInternalApi
 fun runBenchmarks(name: String, @Suppress("unused") args: Array<out String>, declareAndExecuteSuites: (SuiteExecutorBase) -> Unit) {
     val arguments = wasiGetArguments()
-    check (arguments.size >= 3) { "Unexpected number of arguments: ${arguments.size}"}
-    val configPath = arguments[2]
 
-    val executor = when (arguments.size) {
-        3 -> {
+    val executor = when {
+        arguments.any { it == "startAll" } -> {
```
fzhinkin: The number of arguments changed, and it is no longer makes sense to distinguish different cases by it?

igoriakovlev:
I have added single way to these modes argument parsing and building.

fzhinkin (approved):

PR #381, opened 2026-06-29
set mainFile as fileToExecute for WASI development tasks

alex28sh:


fzhinkin:
@alex28sh, the PR gives zero clues on why the change is needed and what problem it solves. It there's indeed a problem, then the change should be accompanied by the test.

alex28sh:
The problem was that for WASI benchmark runners customized by the user, development task ran optimized wasm binary, instead of development one. Will look now, if I can add a test for that

alex28sh:
Added the test for checking, which tasks were run, - for Development benchmarks no optimized compilation tasks should run

fzhinkin:
@alex28sh, thank you!

fzhinkin:
I'll merge the PR once CI will start felling better

fzhinkin (approved):

PR #385, opened 2026-07-23
Fail fast when reading klib with incompatible `abi_version`

radoslaw-panuszewski:


fzhinkin:
@radoslaw-panuszewski, just for the record:
- if the plugin can't load klib, then the issue is likely somewhere else - the plugin depends on the compiler-embeddable's version corresponding to the KGP version from the project (https://github.com/Kotlin/kotlinx-benchmark/blob/77f1ae0144ea7abcc67dd6b5531339c43faa6f1e/plugin/main/src/kotlinx/benchmark/gradle/internal/BenchmarkDependencies.kt#L29)
- this part will be rewritten soon anyway, either to using kotlinx-metadata-klib, or KSP

radoslaw-panuszewski:
Hey @fzhinkin 

> if the plugin can't load klib, then the issue is likely somewhere else

Yep, the real issue is that `kotlin-native/performance` uses local snapshot of K/N and bootstrap version of KGP at the same time. I proposed a fix for that -> https://github.com/JetBrains/kotlin/pull/6958

> this part will be rewritten soon anyway, either to using kotlinx-metadata-klib, or KSP

I thought it would make sense to add this quick fix for the current implementation, just to fail fast. But if you prefer not to merge it, I'm ok with that 😉

radoslaw-panuszewski:
I'm closing this draft for now, let's wait for the either the KSP or kx-metadata solution 😉

PR #386, opened 2026-07-24
pass engineArguments to SpawnBenchmarkExecutor

alex28sh:


runtime/wasmJsMain/src/kotlinx/benchmark/wasm/WasmPerProcessExecutors.kt
```diff
@@ -2,6 +2,14 @@ package kotlinx.benchmark.wasm
 
 import kotlinx.benchmark.*
 
+private fun execArgv(): String =
+    js("process.execArgv.join(' ')")
```
igoriakovlev: Maybe here (and other places) to return JsArray<JsString>?

igoriakovlev (changes requested):

runtime/wasmJsMain/src/kotlinx/benchmark/wasm/WasmPerProcessExecutors.kt
alex28sh: @igoriakovlev Is that what you meant?

As I understand, sometimes it eliminates some wrapping/unwrapping, right?

There is a similar function in NodeJsEngineSupport
```
private fun nodeJsArguments(): String =
    js("process.argv.slice(2).join(' ')")
```

but it's used only for
```
override fun arguments(): Array<out String> =
    nodeJsArguments().split(' ').toTypedArray()
```
that comes from common abstract method - so, as for me, no sense in changing anything there
 
 Otherwise, if changing into 
```
 private fun nodeJsArguments(): JsArray<JsString> =
    js("process.argv.slice(2)")
```
We will get
```
    override fun arguments(): Array<out String> =
        nodeJsArguments().toArray().map { it.toString() }.toTypedArray()
```
That kinda looks weird

alex28sh: Seems like, joining then splitting is redundant here?

refactored in the latest commit into the following
```
private fun nodeJsArguments(): JsArray<JsString> =
    js("process.argv.slice(2)")
```

```
override fun arguments(): Array<out String> {
    val argv = nodeJsArguments()
    return Array(argv.length) { argv[it]!!.toString() }
}
```

igoriakovlev (approved):

PR #389, opened 2026-08-13
Fixed formatter issue caused by changed WASI Double.toString implementation

igoriakovlev:
Kotlin Wasm will change soon Double and Float stringifiers. This will trigger test TextBenchmarkReportFormatter failure. This is a fix to have tests passing.

### DIFF
--- a/runtime/wasmWasiMain/src/kotlinx/benchmark/Utils.kt
+++ b/runtime/wasmWasiMain/src/kotlinx/benchmark/Utils.kt
@@ -36,10 +36,11 @@ private fun Double.toPlainString(): String {
     val mantissa = unsigned.substringBefore("e")
     val exponent = unsigned.substringAfter("e").toInt()
 
-    val integerPart = mantissa.substringBefore('.')
-    val fractionPart = mantissa.substringAfter('.', "")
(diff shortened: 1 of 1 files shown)


runtime/wasmWasiMain/src/kotlinx/benchmark/Utils.kt
```diff
@@ -36,8 +36,8 @@ private fun Double.toPlainString(): String {
     val mantissa = unsigned.substringBefore("e")
     val exponent = unsigned.substringAfter("e").toInt()
 
-    val integerPart = mantissa.substringBefore('.')
-    val fractionPart = mantissa.substringAfter('.', "")
+    val integerPart = mantissa.substringBefore('.').trimStart('0')
+    val fractionPart = mantissa.substringAfter('.', "").trimEnd('0')
 
     val digits = "$integerPart$fractionPart"
```
fzhinkin, line 42: @igoriakovlev, `this.toString().lowercase()` will never return "0.0e...", right?

igoriakovlev: I dont think so. But why do you ask?

fzhinkin (approved):

PR #390, opened 2026-08-18
Remove usage of `KlibMetadataFactories`

ivandev0:
This PR is part of the fix for [KT-87387](https://youtrack.jetbrains.com/issue/KT-87387). Kotlin compiler team will remove `KlibMetadataFactories` and all descriptor factories. Right now the only known external user of these factories is `kotlinx-benchmark` project. The easiest way to fix it – inline all the usages. Later on, all such usages should be fixed by [KT-82882](https://youtrack.jetbrains.com/issue/KT-82882).

I did this change in separate digestible commits. Please tell me if you want to squash then before merging.


fzhinkin (approved):
The change looks good, thanks for opening the PR!

fzhinkin:
Just for the context: while the plan to reimplement the plugin to avoid using internal compiler API is still there, we agreed that this particular change will unblock the compiler team as the plugin rewrite won't happen overnight.

PR #395, opened 2026-08-26
Use KSP instead of the Kotlin compiler embeddable's API

### LINKED ISSUE #396: Avoid using legacy and deprecated compiler API to analyze klibs
The Gradle plugin uses several legacy compiler APIs there were deprecated long time ago and will be removed in 2.5.0.

Related issues in YT are:
- [KT-82882](https://youtrack.jetbrains.com/issue/KT-82882)
- [KT-87387](https://youtrack.jetbrains.com/issue/KT-87387)

The plugin should be reimplemented to avoid using such APIs and rely only on public and stable APIs.

fzhinkin:
Reimplemented the plugin to rely on KSP for multiplatform sources analysis instead of analyzing compiled klibs using internal Kotlin compiler API that is about to be removed in Kotlin 2.5.0.

Plugin uses KSP as a compile-only dependency and the actual dependencies are added dynamically in runtime and utilized only by workers.

Source-generator is using standalone KSP entry point and the way it is initialized is "inspired" by the corresponding KSP Gradle tasks.


fzhinkin:
Generator's rewrite is based on changes contributed by @projedi: https://github.com/Kotlin/kotlinx-benchmark/compare/master...alshabalin/ksp

fzhinkin:
@shanshin, PTAL at the Gradle-specific part of the change.

lppedd:
Does this impose some additional dependency requirement on consumers, or is it a totally transparent change?

fzhinkin:
@lppedd, it's totally transparent to users.

shanshin (approved):
lgtm

PR #399, opened 2026-09-01
Update minSupportedGradleVersion to 8.14.5

Egorand:
Follow-up to d69b7d162027e4ea4ba24dc496fdf16ded8a8857 that sets minSupportedGradleVersion to 8.14.5 to be compatible with the upcoming Kotlin 2.5.0 release. Also updates Gradle version in integration tests so they can run against Kotlin 2.5.0.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -12,7 +12,7 @@ gradle-pluginPublish = "2.1.1"
 ksp = "2.3.11"
 
 # Note: This version can be overridden by passing `-Pmin_supported_gradle_version=<version>`
-minSupportedGradle = "8.0"
+minSupportedGradle = "8.14.5"
 minSupportedKotlin = "2.2.0"
 
 [libraries]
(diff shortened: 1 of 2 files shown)


Egorand:
@fzhinkin - I'm not sure this is the best approach, I understand we might not want to bump `minSupportedGradleVersion` for consumers who don't yet plan to upgrade to Kotlin 2.5.0. Please let me know if you have thoughts on how to bump the version for integration tests only so the internal user projects TeamCity builds pass!

fzhinkin:
@Egorand, thanks for opening the PR. I think, only `gradle/libs.versions.toml` needs to be updated here. I'll update TC config and one of the affected tests in the `master` branch, merge it to kc-dev and then rebase this branch.

Egorand:
> @Egorand, thanks for opening the PR. I think, only `gradle/libs.versions.toml` needs to be updated here. I'll update TC config and one of the affected tests in the `master` branch, merge it to kc-dev and then rebase this branch.

Awesome, thank you!

PR #402, opened 2026-09-03
KT-84955: Remove usages of macosX64 target due to deprecation

mMaxy:


fzhinkin:
For the context: deprecation of some native targets made them unavailable for cross-compilation and that blocks Kotlin bootstrap version update.

PR #406, opened 2026-09-11
Fix empty lines in (S)CSV native benchmark reports generated on Windows

### LINKED ISSUE #405: CSV reports generated on Windows (mingwX64) have empty lines between actual report lines
Steps to reproduce:
- download the reproducer
- run `\.gradlew.bat benchmark` on a Windows machine
- inspect the generated CSV report in `.\build\reports\benchmarks\main`

Expected result: the report is a CSV file without empty lines.
Actual result: the report contains empty lines between each report line.

[reproducer](https://github.com/user-attachments/files/32098023/NativeRandomBenchmark.zip)

nikolaykasyanov:
When writing to a file opened by `fopen(..., "w")`, on Windows each `LF` automatically becomes `CRLF`, the fix is to use `"wb"` instead.

Fixes #405.


runtime/nativeMain/src/kotlinx/benchmark/Utils.kt
```diff
@@ -22,7 +22,8 @@ internal actual fun Double.format(precision: Int, useGrouping: Boolean): String
 }
 
 internal actual fun String.writeFile(text: String) {
-    val file = fopen(this, "w")
+    // Preserve line endings: Windows text mode would turn CSV's CRLF into CRCRLF.
+    val file = fopen(this, "wb")
```
nikolaykasyanov, line 26: Open question: do we want all files (except (S)CSV ones) created by benchmarks to have Unix line endings even on Windows or not?

nikolaykasyanov: One option is adding a `preserveLineEndings: Boolean = false` flag to `String.writeFile` to limit the impact to (S)CSV generation only.

fzhinkin: I don't see an issue with using the same line endings across all platforms. As long as there's no argument for using CRLF on Windows, I'd keep things simple.

fzhinkin (approved):

PR #410, opened 2026-09-15
Bypass IJ platform's limit on max file size

### LINKED ISSUE #408: Large source files are ignored by the plugin for Native, Js and Wasm source sets
After migration to KSP, large source files are no longer analyzed. Apparently, there's some file size limit somewhere in PSI and it could be increased by setting `idea.max.intellisense.filesize` property.

fzhinkin:
ATM, there's no better way to bypass IJ Platform's limitation on the file size one can read via PSI (and KSP internally use it).

Closes #408


fzhinkin:
(TODO(filipp): file a bug against KSP)

plugin/main/src/kotlinx/benchmark/gradle/internal/generator/BenchmarkSourceGenerator.kt
```diff
@@ -65,6 +65,9 @@ internal object BenchmarkSourceGenerator {
                 backend = if (platform == Platform.WasmBuiltIn) "WASM" else "JS"
             }.build()
         }
+        // KSP uses AA and IJ Platform API, the latter has a file size limit.
+        // Large files are silently ignored and are not analyzed. This property "unset" the limit.
```
nikolaykasyanov:
```suggestion
        // Large files are silently ignored and are not analyzed. This property "unsets" the limit.
```

nikolaykasyanov (approved):

plugin/main/src/kotlinx/benchmark/gradle/internal/generator/BenchmarkSourceGenerator.kt
```diff
@@ -65,6 +65,9 @@ internal object BenchmarkSourceGenerator {
                 backend = if (platform == Platform.WasmBuiltIn) "WASM" else "JS"
             }.build()
         }
+        // KSP uses AA and IJ Platform API, the latter has a file size limit.
+        // Large files are silently ignored and are not analyzed. This property "unsets" the limit.
+        System.setProperty("idea.max.intellisense.filesize", "-1")
```
whyoleg, line 70: _Thinking out loud._

Currently, kotlinx-benchmark runs KSP in an [isolated class loader](https://github.com/Kotlin/kotlinx-benchmark/blob/f65212cbd41506cf5c962a681266555d2af9ff41/plugin/main/src/kotlinx/benchmark/gradle/SourceGeneratorTask.kt#L73), which means that the same JVM is used by both Gradle and its spawned workers.
So setting the property here, even in a worker, means that it will be set for the whole Gradle process. (Right?)
Not sure if this will cause any issues, but it still means it might affect other consumers of the IJ platform (and thus AA consumers) in the same Gradle process when the benchmarks task is executed. As well as for any other tasks that will be executed afterward, since the Gradle process is a daemon.

_Though I might be wrong somewhere here. :)_

fzhinkin: Right, thanks! It seems like the only isolated solution here is to spawn a separate process that'll analyze source and all that. Which might not be a bad idea.

whyoleg: Yeah. Be aware that Gradle's `processIsolation` should be used with care, as it will provide kotlin stdlib in its classpath automatically, and as KSP brings AA or some dev version, as well as it requires the stdlib of the same dev version, you need to be careful, as there could be compatibility issues...

Related Gradle issues:
- https://github.com/gradle/gradle/issues/34442
- https://github.com/gradle/gradle/issues/33987

And a "meta" issue in Dokka about different ways to overcome problems with Gradle workers - https://github.com/Kotlin/dokka/issues/4215 - might be useful. :)
Just so that you know, how good the process isolation is.

Good luck? :)

